PR #2, opened 2016-04-15
Typo fix. Add mention of AWS KMS. Make button pop more.

josh-padnick:
Looks like there's a merge conflict right now. May be easier to just manually make these changes on master.


josh-padnick:
Ok, rebased, and ready to merge now. I went with your simplified language on the the What We Do (won't be reflected in commits).

PR #4, opened 2016-07-26
Change package styling

brikis98:
This PR changes how we style packages to make it more obvious each package is clickable and to avoid the awkwardness of expanding one package in a two-column table.

### DIFF
--- a/_data/packages.yml
+++ b/_data/packages.yml
@@ -68,8 +68,3 @@
     Package deploys an EC2 Instance, attaches an EBS volume that persists data between redeploys, configures the
     Instance with a public IP and DNS A record (optional), uses AWS Lambda job to run backup the Instance on a
     cron schedule, and adds an alarm that goes off if a backup job doesn’t run within a scheduled time period.
-- title: "Consulting"
-  image: "logo/grunty-small.png"
-  description: |
-    Don't see a Package that does what you need? We offer general DevOps consulting on an hourly basis. We also offer
-    training in AWS, Terraform, Docker, Packer, and other DevOps technologies and practices.
--- a/_includes/packages.html
+++ b/_includes/packages.html
@@ -1,26 +1,25 @@
-<div class="features">
-  {% for package in site.data.packages %}
-    <section class="expand" ga-on="click" ga-event-category="click" ga-event-action="expand-package">
-      <div class="align-center">
-        {% if package.image %}
-          {% include image.html path=package.image class="icon block mx-auto" %}
-        {% else %}
-          <i class="icon major fa-{{ package.fa_icon }} fa-lg mx-auto"></i>
-        {% endif %}
-        <h3 class="mt2">
-          {{ package.title }}
-        </h3>
-      </div>
-      <div class="display-none collapsed">
-        {% if package.docs_url %}
-          <div class="align-center">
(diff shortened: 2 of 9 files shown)


brikis98:
Screenshots:

screen shot 2016-07-26 at 3 22 27 pm
screen shot 2016-07-26 at 3 22 41 pm

josh-padnick:
This is a really cool look. I'm going to load it locally and play around.

josh-padnick:
This looks great. Some minor suggestions:

- Should we change "Welcome" in the nav to "Gruntwork"?
- Should we change each package's background color when expanded to `rgb(132, 57, 105)` or similar color? Here's a preview:
   image
- The technologies look great and add a lot of credibility to the site. For now I think it's good enough but something to consider for our next iteration:
   - Currently all the technologies cut across categories: cloud provider, language, OS, etc. It'd be nice if we explicitly listed "Supported Cloud Providers: AWS Logo" and "Supported Langauges: Node, Ruby, Java, Scala, etc." Not sure where Docker would go.
   - It'd be nice to clarify that this list isn't fixed and to add a statement that if they don't see a tech they want, contact us and maybe we can discuss adding support for it.

Overall, I feel like this is a significant improvement on the website from a few days ago. Nice!

josh-padnick:
Some other feedback:

- Should we add `border:1px solid #999` to the blog images? Example:

   image

josh-padnick:
Finished reviewing. This is looking great! Thanks for taking the lead on this.

brikis98:
> Should we change "Welcome" in the nav to "Gruntwork"?

I could go either way on this one. I kind of like "Welcome", but I can see how the menu is missing our branding to some extent.

> Should we change each package's background color when expanded to rgb(132, 57, 105) or similar color

I agree that the design could use a bit more contrast, but this color change doesn't look quite right either. Color schemes are a PITA to get right. I'd rather not dive into that rabbit hole at the moment and stick with the defaults from the html5up template.

> Currently all the technologies cut across categories: cloud provider, language, OS, etc. It'd be nice if we explicitly listed "Supported Cloud Providers: AWS Logo" and "Supported Langauges: Node, Ruby, Java, Scala, etc." Not sure where Docker would go.

Yea, some sort of categories will be good as the number of techs grow.

> It'd be nice to clarify that this list isn't fixed and to add a statement that if they don't see a tech they want, contact us and maybe we can discuss adding support for it.

Added.

> Should we add border:1px solid #999 to the blog images?

Hm, not sure it adds much. Is this for contrast?

brikis98:
Merging!

josh-padnick:
Sounds good on everything.

> Should we add border:1px solid #999 to the blog images?
> Hm, not sure it adds much. Is this for contrast?

Yeah, the fuchsia background clashes a little bit with the images. I thought a common border would clean it up a little.

brikis98:
> Yeah, the fuchsia background clashes a little bit with the images. I thought a common border would clean it up a little.

How about a thicker, white border?

screen shot 2016-07-28 at 12 50 44 am

josh-padnick:
I think these looks great. I might play around with a more gray-ish white to see if that takes some emphasis on the image, but otherwise, makes the blog part look a little more polished.

brikis98:
OK, the border is now live.

PR #5, opened 2016-08-29
Add blog post link to Josh's AWS presentation.

josh-padnick:
@brikis98 Let me know if this looks good to you!

### DIFF
--- a/_data/blog_posts.yml
+++ b/_data/blog_posts.yml
@@ -1,3 +1,12 @@
+- title: "A talk on the new AWS Application Load Balancer, updates to ECS, and Kinesis Analytics"
+  url: "https://blog.gruntwork.io/a-talk-on-the-new-aws-application-load-balancer-updates-to-ecs-and-kinesis-analytics-abb599cb3cb8#.vhbek5ce2"
+  thumbnail_url: "https://i.imgur.com/KkniqPD.png"
+  date: "08/29/2016"
+  tags:
+    - AWS 
+    - Infrastructure
+  excerpt: |
+    Josh presented to the Official AWS Phoenix Meetup group on three new services and updates released by AWS.
 - title: "Add Automatic Remote State Locking and Configuration to Terraform with Terragrunt"
   url: "https://blog.gruntwork.io/add-automatic-remote-state-locking-and-configuration-to-terraform-with-terragrunt-656a57565a4d"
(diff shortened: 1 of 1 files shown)


_data/blog_posts.yml
```diff
@@ -1,3 +1,12 @@
+- title: "A talk on the new AWS Application Load Balancer, updates to ECS, and Kinesis Analytics"
+  url: "https://blog.gruntwork.io/a-talk-on-the-new-aws-application-load-balancer-updates-to-ecs-and-kinesis-analytics-abb599cb3cb8#.vhbek5ce2"
+  thumbnail_url: "https://i.imgur.com/KkniqPD.png"
```
josh-padnick: If there's a better place to upload this, let me know. I didn't see a way to upload arbitrary images on Medium separate from the post itself.

brikis98: Well, if you add an image into the Medium post itself, it will a) use it as a thumbnail on the [blog.gruntwork](https://blog.gruntwork.io/) page, which will look much better and b) you'll be able to use that image's URL here.

brikis98:
Changes lgtm. Merge at will.

_data/blog_posts.yml
brikis98: Actually, you *might* be able to use the SlideShare deck itself as a thumbnail: https://help.medium.com/hc/en-us/articles/215680047-Featured-image

Select it in the post and hit `CMD+OPTION+8`. Not exactly intuitive or discoverable, is it?

josh-padnick: Nice find! Unfortunately, I don't think they support SlideShare's quite right. Although I get a message box confirming the SlideShare is now the featured image the green border doesn't turn red and the thumbnail never actually adjusts.

It doesn't look like it's possible to set a thumbnail of an image not in the story?

brikis98: No, looks like only for images in the story. You could use a random splash image at the top. Perhaps the diagram that shows the ALB and ELB comparison?

josh-padnick: Ugh, Medium takes a single image and places it in 3 different proportions, cutting it off at arbitrary points, depending on where you're viewing the article or headline. I added something that's workable but doesn't look that great.

I'm going to merge this so we have it live and I can move on, but feel free to tinker with this if you have other ideas. Ultimately, I had to format my image to match the proportions in the pages that list multiple articles, but it still doesn't look very good overall.

brikis98: Looks good enough to me!

PR #16, opened 2017-09-16
Update v2 branch so it works in dev & prod

brikis98:
This PR updates the v2 branch, which is where the redesigned website is (rather than master, so it's not live!), with the following changes:

1. Update Jekyll so it can compile less automatically and do hot reload, rather than requiring a separate manual compilation step. This requires a custom Jekyll plugin, so we can no longer use GitHub pages. Therefore, we need to move the site to S3 + CloudFront, which is good to do anyway so we can have HTTPS.

1. Add Terraform code that uses our own modules to create an S3 bucket + CloudFront distro. Note that the Route 53 entry is still managed in the Phoenix DevOps account. We should move that out at some point. 

1. Update the Docker container to move off alpine since it does not work with rubyracer, which is necessary for using less with Ruby.

1. Install [s3_website](https://github.com/laurilehmijoki/s3_website) in the Docker container as a way to push the static content to S3. This neat binary does a diff before uploading to ensure only changed files are uploaded and takes care of setting cache headers and gzip settings for the files too. Pushing live is now a single command: `docker-compose run web s3_website push`.

1. Remove unused files.

1. Fix permalinks.

1. Update README

All the changes are now deployed at: https://redesign.gruntwork.io/


s3_website.yml
```diff
@@ -0,0 +1,4 @@
+s3_bucket: www.gruntwork.io
+cache_control: max-age=600
+gzip: true
+concurrency_level: 4
```
brikis98: `s3_website` is a very handy tool. We should add it to usage-patterns so customers can use it too!

terraform/main.tf
```diff
@@ -0,0 +1,77 @@
...
+  should_redirect_all_requests = true
+  redirect_all_requests_to     = "${module.static_website.website_domain_name}"
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# CREATE THE CLOUDFRONT WEB DISTRIBUTION
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "cloudfront" {
+  source = "git::git@github.com:gruntwork-io/package-static-assets.git//modules/s3-cloudfront?ref=v0.0.1"
```
brikis98: I'm very happy to finally be dogfooding our own infra. It was wonderfully simple to get the S3 + CloudFront stuff running.

brikis98:
Merging now so I can start working on the actual content/design changes. Feedback welcome!

PR #22, opened 2017-09-20
Update to new package-static-assets version with multiple domain support

brikis98:


brikis98:
Merging now. Feedback welcome.

PR #26, opened 2017-09-29
Update library links.

josh-padnick:
This commit:
- Adds links to all infra packages, even ones where we don't have a docs
repo. If no docs repo exists, a new slightly different modal is used,
and a small label is next to the link that says "(subscribers only")
- Fixed links to the open source repos. Previously, these forgot the
"https://"

These changes were motivated by the desire to make it easier for
subscribers to browse everything in the library. In the future, I'd like
to add a small checkbox that allows users to stop the modal on every
click.

PR #27, opened 2017-09-29
Tweaks per @brikis98

josh-padnick:
Per @brikis98 feedback in #26


josh-padnick:
The link feels a bit long now, but seems better than just linking the word "subscribers"

PR #29, opened 2017-11-07
Clarify that billing frequency is monthly.

josh-padnick:
In Slack we discussed that:

> We need to clarify that while a customer commits to 12 months we only charge them on a monthly basis. A number of potential customers have emailed and asked if they could pay monthly for the IAC Library and support. This confused us because we already list the price as monthly. But then we noticed something: we say that the "Subscription Period" is "annual."

> We think people interpret it as "you must pay for the whole year up-front." I bet that has scared a number of potential customers off. It's prob a small tweak, but ... we should make it clear we bill monthly, but that the contract is for 1 year.

Rather than engage the design team to tackle this, I made this simple tweak.

### DIFF
--- a/pages/pricing/_infrastructure-as-code-library.html
+++ b/pages/pricing/_infrastructure-as-code-library.html
@@ -39,6 +39,10 @@
           <td>Annual</td>
         </tr>
         <tr>
+          <th>Billing frequency</th>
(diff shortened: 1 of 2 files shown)


josh-padnick:
Pushed live.

PR #33, opened 2017-11-16
Add comparison table to the FAQ

brikis98:


brikis98:
Pushed live.

josh-padnick:
This is great! So cool to see a direct 10x comparison.

I read the rest of the page and wonder if we're coming down too hard on consulting companies. I think what we want to come down on is not all consulting, but consulting that is getting paid to do _generic_ work. Should we update content accordingly? Note that I'm referring to content on the same web page, but not edited in this PR.

brikis98:
> This is great! So cool to see a direct 10x comparison.

It's straight from the exec summary slide deck: https://www.slideshare.net/brikis98/gruntwork-executive-summary-81144627

> I read the rest of the page and wonder if we're coming down too hard on consulting companies. I think what we want to come down on is not all consulting, but consulting that is getting paid to do generic work. Should we update content accordingly? Note that I'm referring to content on the same web page, but not edited in this PR.

Sure, feel free to make some tweaks to make that clearer.

PR #34, opened 2017-11-17
Update Ref Arch to say encryption instead of TLS

brikis98:
End-to-end encryption applies to data in transit and at rest, so it’s not just TLS.


brikis98:
Pushing live now.

PR #37, opened 2017-12-05
Careers tweaks

josh-padnick:


mcalhoun:
Link Broken: Quip: A Counterintuitive System for Startup Compensation

josh-padnick:
@mcalhoun Thanks for the head's up. Fixed! cfd36ba5c9a634d8f1560cac29e46c7a736fc3d2. It will about 10 mins before it shows on the live site.

pages/careers/_open-positions.html
```diff
@@ -3,27 +3,29 @@
...
-      <li>Demonstrated experience building good software.</li>
-      <li>Bonus points for visible open source work.</li>
-      <li>Bonus points for any technical presentations you've done in the past.</li>
-      <li>Bonus points for a blog.</li>
-      <li>Bonus points for something in your profile that makes us smile.</li>
-      <li>To be clear, you don't have to be good at everything. Our most important criteria are:
+      <li>You know how to write code across the stack ("Dev").</li>
+      <li>You have experience running production software ("Ops").</li>
+      <li>You have a strong background in software engineering.</li>
+      <li>Bonus points for:
```
brikis98: My worry is that this list, as presented, will scare off candidates who don't have these. I mean, the other things that could be part of this list are strong previous employers, top-tier schools, and personal references, but again, none of those are required. So it's not obvious to me how much we gain by listing these, versus how much we may lose.

Perhaps instead of listing each item as a separate bullet point

josh-padnick: I guess I don't feel that strongly about it, so if it makes you worried, I'm inclined to remove. Thoughts?

brikis98: Haha, just realized I didn't quite finish typing that previous comment, did I? Sorry, multi-tasking a bit too much today.

I like the idea of providing more info on what we're looking for. In a sense, I want to say that these are the things we consider, but none are required (including, crucially, things like "top-tier school" and "CS degree" that are stupidly required by so many other jobs). However, I'm not sure how to do it in a way that is genuine and candidates won't misinterpret as an intimidating checklist of "must-haves."

So for now, removing is fine, but I'm definitely open to other ways of phrasing this.

josh-padnick: Ok, simplified this in https://github.com/gruntwork-io/gruntwork-io.github.io/pull/37/commits/f22c3c7cdcc02a2bf37853ef417e1ebfac445a9f and pushed it live. Merging now!

PR #41, opened 2018-04-02
Add pricing to nav.

josh-padnick:
I also discovered that our FavIcon was missing so added back our original one.


josh-padnick:
This has already been pushed live!

brikis98:
Nav change LGTM.

I don't think the favicon is missing. We just don't link to it at the root: https://github.com/gruntwork-io/gruntwork-io.github.io/blob/master/_includes/favicon.html

brikis98:
Hm, turns out the browser, when loading non-HTML pages (e.g. if loading an image), will always request it at the root, since non-HTML pages don't have link tags: https://stackoverflow.com/questions/21359342/is-putting-your-favicon-ico-file-in-a-non-root-path-a-bad-idea. Therefore, having it at the root or making the root a redirect is a good idea 👍

PR #48, opened 2018-06-08
IE 10/11 Fix for Flex Columns Direction - Add width 100%

sahilakos:
This should fix the overflow issue on IE for panels.


josh-padnick:
Ah, thanks for the clarification. Ok, merging!

josh-padnick:
I'll deploy the changes so they should be live on our site in about 10 mins.

PR #49, opened 2018-06-22
Add mention of extensive contract review fee.

josh-padnick:
Here's how this renders on the [pricing page](https://gruntwork.io/pricing/):

image

### DIFF
--- a/pages/pricing/_other-services.html
+++ b/pages/pricing/_other-services.html
@@ -13,5 +13,12 @@
         </tr>
       </tbody>
     </table>
+
+    <h3>Miscellaneous Fees</h3>
+    <p><strong>Gruntwork has no hidden fees.</strong> The one fee we apply is a one-time charge of $4,000 for any significant requested
+    changes on our contract to cover the cost we pay to our lawyers to review this. Minor contract change requests are
+    done inhouse and cost nothing.</p>
   </div>
 </div>
+
+
--- a/pages/pricing/_other-services.html
+++ b/pages/pricing/_other-services.html
@@ -14,10 +14,9 @@
       </tbody>
     </table>
 
-    <h3>Miscellaneous Fees</h3>
-    <p><strong>Gruntwork has no hidden fees.</strong> The one fee we apply is a one-time charge of $4,000 for any significant requested
-    changes on our contract to cover the cost we pay to our lawyers to review this. Minor contract change requests are
-    done inhouse and cost nothing.</p>
+    <h2 style="padding-top: 27px;">Legal Fee</h2>
+    <p>The pricing on this page comes with our standard contract. If you request significant changes on that contract, we
+      add a fee of $5,000 to pay our lawyers. Minor contract changes are included at no additional cost.</p>
   </div>
 </div>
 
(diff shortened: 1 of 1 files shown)


pages/pricing/_other-services.html
```diff
@@ -13,5 +13,12 @@ <h2>Other Services</h2>
         </tr>
       </tbody>
     </table>
+
+    <h3>Miscellaneous Fees</h3>
```
brikis98: `h2`

pages/pricing/_other-services.html
```diff
@@ -13,5 +13,12 @@ <h2>Other Services</h2>
...
+    <p><strong>Gruntwork has no hidden fees.</strong> The one fee we apply is a one-time charge of $4,000 for any significant requested
+    changes on our contract to cover the cost we pay to our lawyers to review this. Minor contract change requests are
+    done inhouse and cost nothing.</p>
```
brikis98: Although I like the sentiment, and we really don't hide any fees, we also don't display our enterprise pricing or consulting pricing on the website. Perhaps just keep it simple and say "Legal fee": the pricing on the page comes with our standard contract. If you request *significant* changes on that contract, we add a fee of XXX to pay our lawyers. Minor contract changes...

pages/pricing/_other-services.html
```diff
@@ -13,5 +13,12 @@ <h2>Other Services</h2>
         </tr>
       </tbody>
     </table>
+
+    <h3>Miscellaneous Fees</h3>
+    <p><strong>Gruntwork has no hidden fees.</strong> The one fee we apply is a one-time charge of $4,000 for any significant requested
```
brikis98: I'd make it $5k based on our legal team's estimates.

josh-padnick:
Made the requested changes. I think it sounds better. My one concern though is that how does someone distinguish from us simply passing around our costs versus us exorting a customer into accepting our "unreasonable" terms? For example, suppose we have an especially onerous clause or we offer no indemnity and then claim that a customer who wishes to challenge that will be charged $5k to do so.

I'll experiment with some language to address this point.

brikis98:
What you have there LGTM. I'd add that this is a temporary stop gap. Soon, we'll have self-checkout with an online ToS. So the only people this will affect are those who refuse to sign up online.

mcalhoun:
> Made the requested changes. I think it sounds better. My one concern though is that how does someone distinguish from us simply passing around our costs versus us exorting a customer into accepting our "unreasonable" terms? For example, suppose we have an especially onerous clause or we offer no indemnity and then claim that a customer who wishes to challenge that will be charged $5k to do so.

> I'll experiment with some language to address this point.

They always have the ability to negotiate and then not sign the deal, owing us nothing, right?

josh-padnick:
Thx for the great feedback, guys. Two good points:
- With self-checkout, the customer can both see exactly what our contract is and understand clearly why we need to charge a fee to deviate from it.
- If the customer finds our terms onerous, they can always choose not to sign, though we'll of course make every effort to create a fair contract.

Merging and deploying now!

PR #60, opened 2018-08-09
add hotjar tracking to footer scripts

sahilakos:
@josh-padnick can you approve and deploy. Will finish setup in Hotjar once it's deployed.


josh-padnick:
I'm guessing you just copy & pasted their install script, so LGTM. I'll merge now and confirm here once it's deployed.

josh-padnick:
Deployed!

PR #61, opened 2018-08-18
Remove legal fee.

josh-padnick:
This no longer reflects our current policy, so removing.


josh-padnick:
This has been deployed.

PR #63, opened 2018-08-28
[DO NOT MERGE] Ugly hack to be able to produce ToS in Word Doc format

brikis98:
This PR adds a page that can render our Terms of Service in plain HTML with no extra markup or styling. Combined with Pandoc, I was able to use this to produce a reasonable-looking Word Doc version of our Terms of Service as follows:

```
pandoc -s -r html http://localhost:4000/terms/word -o terms.docx
```

The result in `terms.docx` is passable, though it uses the default Word styling, which is a bit ugly, so I cleaned up the styles manually. If there's a way to clean up the styles automatically, then we can add Pandoc to the Docker image and make it part of the Jekyll build process. 

This PR is not meant to be merged directly, but to be used by @sahilakos as a possible starting point.

@sahilakos See if this helps you at all.

### DIFF
--- a/_data/terms-of-service.yml
+++ b/_data/terms-of-service.yml
@@ -1,5 +1,5 @@
 - category_name:
-  category_id: general
+  category_id: tos
   category_description:
   items:
     - title:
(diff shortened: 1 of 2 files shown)


josh-padnick:
These changes were implemented and merged in #67, so closing.

PR #69, opened 2018-09-27
Basecamp: General Updates - Sept 20, 2018

sahilakos:
https://3.basecamp.com/3989840/buckets/9063434/todolists/1313507914


josh-padnick:
Thanks for submitting this, @sahilakos! Some feedback:

1. On the bootcamp page, when you go to a link like http://0.0.0.0:4000/devops-bootcamp/#day-2, the page successfully opens the right section, but there appears to be a race condition where the page loads, the browser scrolls down to the anchor, and _then_ the section expands, which leave the user in a confusing spot on the page like this:

   [Screen Recording 2018-09-27 at 4.28.46 PM.mov.zip](https://github.com/gruntwork-io/gruntwork-io.github.io/files/2426221/Screen.Recording.2018-09-27.at.4.28.46.PM.mov.zip)

2. Pricing page and Get Started looks good.

sahilakos:
@josh-padnick Should be fixed now, this was a doozie. Took like 2 hours to debug (marked non-billable). The issue was only on firefox. Turns out having the first accordion open by default for Day-1, was causing the direct links positioning to be messed up. So, I removed the open by default class for Day 1, and am now opening it using JS if there's no hash in the URI. Test it out and let me know if it works well.

josh-padnick:
5 weeks later, just verified these changes and they look great. Merging and releasing!

PR #72, opened 2018-10-10
Remove duplicate section in terms.

josh-padnick:


brikis98:
Do you have to update the "date these terms were last changed"?

josh-padnick:
Eh, this is literally a duplicated paragraph, so nothing material changed.

PR #75, opened 2018-10-25
Fix minor typos

equinox:
Remove superfluous "into" and correct spelling of "library".


josh-padnick:
Thanks so much for the fix!

PR #79, opened 2018-11-21
Fix typo in tos

yorinasub17:


josh-padnick:
Thanks, Yori! I'll deploy this now.

PR #87, opened 2018-12-11
Remove third-party calls to ipinfo.io & re-add missing cookie modal styles

sahilakos:
@josh-padnick This PR should take care of the third party calls, and there were some missing styles for cookie modal in the last merge, so added them back. Let me know if there's any further changes to the modal.

### DIFF
--- a/_data/devops-bootcamp.yml
+++ b/_data/devops-bootcamp.yml
@@ -35,7 +35,7 @@ day_three:
     description: A Gruntwork engineer goes through a checklist of questions with your team to see what work you need to do to be ready for prod.
 
   - title: Architecture deployment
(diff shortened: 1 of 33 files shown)


sahilakos:
@josh-padnick @brikis98 I decided to add another commit to this PR based on Rob's task. That was a good find and we should always have trailing slashes on internal links. It avoids unnecessary 302 redirects. Last commit makes changes to a lot of files, but it's just adding trailing slashes on links so it shouldn't affect anything else. I will run another crawl on the site once this is deployed.

josh-padnick:
Thanks, @sahilakos and @robmorgan for helping to make this happen! Merging now and deploying live shortly.

robmorgan:
Thanks for fixing this so quickly @sahilakos 👍

PR #88, opened 2018-12-13
Add Amy and Monica to About page

sahilakos:
Amy and Monica added to about page. No bio for Bijou yet.


sahilakos:
@brikis98 Done.

PR #89, opened 2018-12-26
re-add wrongly deleted styles in fba573f0c7e7c685a918c541238d759cc34f…

sahilakos:
@josh-padnick This should take care of the issues with wrongly deleted styles. I think something got messed up in my local that it overwrote so many styles on Nov 20. Should be fine now.

### DIFF
--- a/assets/css/app.less
+++ b/assets/css/app.less
@@ -82,6 +82,11 @@ p + .btn {
 }
 .btn-primary-hollow {
     border: 1px solid #13cfe7;
+    &:hover{
+        background: #9c9c9c;
+        background: -webkit-linear-gradient(-45deg, #3c3c3c, #5e5e5e);
+        background: linear-gradient(135deg, #3c3c3c, #5e5e5e);
+    }
 }
 
 // Footer buttons styles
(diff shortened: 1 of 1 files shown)


brikis98:
What do these deleted styles do? Hard to tell from just looking at the code.

josh-padnick:
> What do these deleted styles do?

It's a fair question. Most of these are optimizations I made to the checkout experience and Terms & Conditions. E.g. `code` makes the formatting of all source code look more appropriate for our site. 

Ideally, we'd clearly comment each of them, but honestly this is what was there before and I'm not sure the additional clarity is worth the commenting effort. My vote is to merge as is.

brikis98:
No need to comment. It's just that usually, the styling changes go with functionality changes, so it's clearer what is being added/updated. In this case, they are standalone, so it's more confusing. I'm OK merging and releasing as-is.

josh-padnick:
Oh, the context here is that [this commit](https://github.com/gruntwork-io/gruntwork-io.github.io/commit/fba573f0c7e7c685a918c541238d759cc34f00ec#diff-149baf93473d97ec899cef923e605e48) accidentally removed a bunch of styles and Sahil is restoring these. Great, I'll merge and deploy now!

PR #94, opened 2019-02-14
Add user time zone to contact us form.

josh-padnick:
The overall goal here is to allow Salesforce.com to automatically assign leads to either Rob or Josh based on the timezone the user submits their inquiry from. I originally thought locale was the way to do this, and then I realized that Javascript directly exposes the user's timezone offset, so I went that route.

Now when the contact us form is submitted, it will include a field "User UTC Timezone Offset." That field has also been added to Salesforce, and I've created a [lead assignment rule](https://help.salesforce.com/articleView?id=customize_leadrules.htm&type=5) to automatically assign leads as follows:

- If the user's timezone is > UTC-1 and < UTC+11, it will be assigned to Rob.
- If the user's timezone is > UTC+11 or < UTC-1, it will be assigned to josh

UTC timezones worldwide range from UTC+12 to UTC-12, with the USA ranging from UTC-11 to UTC-4. See https://www.worldtimezone.com/ for a visual map of them. This assignment system should be active within the hour.

### DIFF
--- a/assets/js/contact.js
+++ b/assets/js/contact.js
@@ -1,33 +1,16 @@
 /**
- * Fetch the user's timezone (expressed in UTC) and write it to the contact us form submission. This is useful because
- * we'll use the user's timezone to assign inbound leads.
+ * Fetch the user's timezone offset relative to UTC and write it to the contact us form submission. This is useful because
+ * we'll use an Assignment Rule in Salesforce.com to automatically assign leads based on the offset value.
  */
 
 (function () {
-
-  // Given an offset in mins, return a well-formatted time zone string.
-  //
-  // Example:
-  //   formatOffset(420)
-  //   returns "UTC+7"
-  function getUtcTimezone(offsetMins) {
-    var offsetHrs = offsetMins / 60;
-
-    if (offsetHrs > 0) {
-      return "UTC+" + offsetHrs;
-    } else if (offsetHrs == 0) {
-      return "UTC(0)";
-    } else {
-      return "UTC" + offsetHrs;
-    }
-  }
 
   // Per https://stackoverflow.com/a/34602679, the main limitation of using Javascript's builtin getTimezoneOffset() to
   // calculate Timezone is that daylight saving rules may change on several occasions during a year. But this doesn't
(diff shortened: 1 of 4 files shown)


pages/contact/_input-hidden.html
```diff
@@ -0,0 +1,5 @@
+<div class="form-group">
+  <div class="row">
+      <input id="{{ include.id }}" name="{{ include.id }}" type="hidden" class="form-control">
+  </div>
+</div>
```
brikis98: If it's hidden, do we need the wrapper `div`s?

assets/js/contact.js
```diff
@@ -0,0 +1,21 @@
...
+  // matter for our purposes, where we only need to know the approximate time zone to assign the lead to the right person.
+  var offsetMins = new Date().getTimezoneOffset();
+
+  // Why multiply by -1? Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset#Description:
+  // The time-zone offset is the difference, in minutes, from local time to UTC. Note that this means that the offset is
+  // positive if the // local timezone is behind UTC and negative if it is ahead. For example, for time zone UTC+10:00
+  // (Australian Eastern Standard Time, Vladivostok Time, Chamorro Standard Time), -600 will be returned.
+  var offsetHrs = (offsetMins / 60) * -1;
+
+  $('input#user-utc-timezone-offset').val(offsetHrs);
```
brikis98: The `input` bit is probably optional since you're targeting directly by ID.

pages/contact/_input-hidden.html
josh-padnick: Yeah, I thought about that but I didn't know if there was other Javascript relying on this somehow. I can try removing and see if that still works.

assets/js/contact.js
josh-padnick: Fair enough, but I'd rather be more descriptive to the reader so they know this is an input we're setting the value of.

josh-padnick:
PR feedback responded to. Merging and deploying!

PR #97, opened 2019-02-21
Interim Pricing Update

josh-padnick:
While we get ready to launch our official new pricing, this is an update in the interim so that our new pricing is now mostly reflective of what's on our website. Here are the specific updates made:

1. **Support Page**

The support page has been modified to explicitly mention Community Support and to present Dedicated Support as a paid option:

image

2.  **Pricing Page**

The pricing page has been updated as follows:

- Gruntwork Subscription is now $795/month for 5 users + $79/month for each additional user
- Page defaults to Gruntwork Subscription without Dedicated Support
- Gruntwork Subscription + Dedicated Support is now $2,745/month for 10 users + $79/month for each additional user

image

image

3. **Checkout Page**

- The checkout page now defaults to no Dedicated Support
- Community Forum has been replaced with Community Support
- Updated pricing is reflected

4. **Limited Refactoring**

I also made page-specific CSS updates one notch easier to manage. Feedback is welcome!


_data/pricing.yml
```diff
@@ -1,26 +1,44 @@
...
+    price: 79
+    checkout:
+        - price: 79
+          url: /checkout/?plan=2
+    details: |
+      per <span class="help-text light" title="A user is any human or machine that makes use of any of Gruntwork's services (Infrastructure as Code Library, DevOps Training Library, Dedicated Support)" data-toggle="tooltip">user</span> per month
+
+dedicated_support:
+  tier1:
+    title: Up to 10 users
```
brikis98: Why 10 here but 5 there? It makes the price points hard to compare...

_data/pricing.yml
```diff
@@ -1,26 +1,44 @@
...
+    users_max: 5
```
brikis98: ?

_data/pricing.yml
```diff
@@ -1,26 +1,44 @@
...
+    checkout:
+      - price: "2,745"
+        url: /checkout/?plan=1&addon=dedicated-support
+    details: |
+      monthly flat fee
+
+  tier2:
+    title: Users 11 - 50
+    users_max: 50
+    price: 79
```
brikis98: Not 99?

pages/checkout/_subscription.html
```diff
@@ -4,7 +4,7 @@
       <div class="col-md-8 panel panel-default panel-light checkout-panel-left">
         <div class="panel-body">
           <div class="row checkout-users">
-            <div class="col-sm-8 col-sm-offset-2">
+            <div class="col-sm-10 col-sm-offset-1">
```
brikis98: Why make this wider?

pages/support/_community-support.html
```diff
@@ -0,0 +1,11 @@
+<div class="row row-page-header">
+  <div class="col-xs-12">
+    <h2>Community Support</h2>
+    <div class="subheader">Included with All Gruntwork Subscriptions</div>
+    <p>
+      All Gruntwork customers receive access to our community-only chat room in Gitter, where you can engage with other
+      Gruntwork customers and get occasional support from Gruntwork engineers.
```
brikis98:
```suggestion
      Gruntwork customers and Gruntwork engineers.
```

"occasional support" is a bit wishy-washy. I understand the goal is to make dedicated support seem more valuable, but I don't know that this is the way to do it.

assets/css/app.less
```diff
@@ -2,2444 +2,16 @@
...
 @import "user10/variables";
 @import "app-variable-overrides";
 @import "user10/utilities";
 @import "vertical-carousel";
 @import "lity";
 
+// Import global styles
+@import "global";
 
-
```
brikis98: What's the advantage of moving this from `app.less` to `global.less`?

_data/pricing.yml
josh-padnick: 10 users for $2,745/month just didn't seem like a good enough value to me. Do you think we should make the base library $795 for 10 users to keep them the same?

brikis98: You mean 5 users for $2,745?

_data/pricing.yml
josh-padnick: Whoops, that was a mistake. Fixed in 8a577132b7bdc2c67a34784582cb2dd5847a08e9.

_data/pricing.yml
josh-padnick: Again, I just felt it was getting too expensive. At this point, we're at 10 users for $2,745/month. I really want our pricing to seem like a no-brainer value and I like the idea that for less than $1,000/year, your dev can access the entire subscription and support.

pages/checkout/_subscription.html
josh-padnick: "1 - 10" is wider than "1 - 5". Without the widening, this text was wrapping.

pages/support/_community-support.html
josh-padnick: Ha, I think I took that language from a sentence you wrote. But sure, I'll update to your language. I'll also clarify that no SLA is provided in this tier.

_data/pricing.yml
josh-padnick: Sorry, yes, I mant 5 users for $2,745.

_data/pricing.yml
brikis98: Where is the $1,000/year number from?

brikis98: Oh, I see, the $79 * 12 vs $99 * 12.

pages/checkout/_subscription.html
brikis98: Ah, gotcha

_data/pricing.yml
brikis98: Alright, sure. I have no compelling evidence one way or the other.

Setting both to 10 users makes sense then.

However, sanity check: will this cause any unexpected behavior amongst existing customers? Will those paying for, say, 9 users, ask to have their price reduced?

josh-padnick: Hmm, possibly. This new, interim pricing is less expensive in terms of the number of users, but much more expensive for Dedicated Support. And almost all the customers with a large number of users also have dedicated support, which would mean that their current pricing is still probably more favorable for them.

That being said, if a customer is moved to the point of saying something about the new pricing, I'm inclined to allow them to switch to it in most cases. Any lingering concerns with that?

assets/css/app.less
josh-padnick: I wasn't able to get the code to reference a file other than `app.less`, so I kept `app.less` as the main point of entry and then had it point to all the "global" settings in `global.less`. This set the stage for having less styles in `global.less` and more in page-specific LESS files.

_data/pricing.yml
grealish: just spotted this in my timeline, 
I would always offer the new pricing to a current customers, if it's cheaper, then it should be on the basis of a "new contract" so 12-month commitment if they wish to fallback, they go back to old arrangement with a 12-month commitment on the "old pricing" 
either way, a change of subscription should mean 12 months commitment, 
Many mobile providers do it that way, as to break even on the dialog required they need a contract extension

josh-padnick: @grealish That sounds reasonable: Any customer that wants to adopt the new pricing can at any time, and starts a new contract at that time.

PR #103, opened 2019-03-08
Change when Cookie Notice is Rendered

drmzio:
The cookie notice will render above the Intercom bubble. Here is what happens behind the scenes:

1. Intercom script is creating new asynchronous script.
2. This can can cause a long delay if the user’s internet is slow.
3. When script is loaded, we can detect it and load the cookie notice right after.
4. Cookie notice should show up on top (above) of the Intercom bubble.

This PR also contains the new position of the Price View on the checkout page for mobile viewports.


_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
 </script>
 <!-- Intercom -->
 <script>(function(){var w=window;var ic=w.Intercom;if(typeof ic==="function"){ic('reattach_activator');ic('update',intercomSettings);}else{var d=document;var i=function(){i.c(arguments)};i.q=[];i.c=function(args){i.q.push(args)};w.Intercom=i;function l(){var s=d.createElement('script');s.type='text/javascript';s.async=true;s.src='https://widget.intercom.io/widget/rzcjacgd';var x=d.getElementsByTagName('script')[0];x.parentNode.insertBefore(s,x);}if(w.attachEvent){w.attachEvent('onload',l);}else{w.addEventListener('load',l,false);}}})()</script>
+<script>
+  (function ($) {
+    var cookieInnerHtml = '<div><p>By using this website you agree to our <a href="/cookie-policy/">cookie policy</a></p><button id="cookieModalClose" class="btn btn-primary">OK</button></div>';
+
+    var getCookiebyName = function (name){
+      var pair = document.cookie.match(new RegExp(name + '=([^;]+)'));
```
brikis98: Can you add a comment explaining what you're doing here and an example of what you're trying to parse? It's possible to reverse engineer it from the regex, but why make maintainers do extra work?

_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
...
+      return !!pair ? pair[1] : null;
```
brikis98: Check `pair.length >= 2`?

_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
...
+      $cookieModal.attr('id', 'gruntyCookie');
+      $cookieModal.css('z-index', '2147483647');
+      $cookieModal.html(cookieInnerHtml);
+
+      $(document).on('click', '#cookieModalClose', function (event) {
+        setCookie('GruntyCookie', '1', 365);
+        $cookieModal.hide();
+      });
+
+      //$(document, 'body').append($cookieModal);
```
brikis98: Remove?

_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
...
+          return;
+        }
+        // Intercom has booted
+        // Remove the interval
+        clearInterval(cookieInterval);
+        initCookie();
+      }, 1000);
+    } else {
+      initCookie();
+    }
```
brikis98: Indentation doesn't look right on the lines above

_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
...
+      //$(document, 'body').append($cookieModal);
+      $('body').append($cookieModal);
+    };
+
+    if (window.hasOwnProperty('Intercom')) {
+      // Perform multiple checks on Intercom.booted property
+      var cookieInterval;
+      cookieInterval = setInterval(function () {
+
+        if (typeof window.Intercom.booted === 'undefined') {
```
brikis98: Is this a public API we can rely on? Add link to relevant docs in comments?

_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
...
+      $(document).on('click', '#cookieModalClose', function (event) {
+        setCookie('GruntyCookie', '1', 365);
+        $cookieModal.hide();
+      });
+
+      //$(document, 'body').append($cookieModal);
+      $('body').append($cookieModal);
+    };
+
+    if (window.hasOwnProperty('Intercom')) {
```
brikis98: I assume this means the cookie code MUST go after the Intercom code. If so, can you add a comment at the top of the cookie code explaining this?

_includes/scripts.html
```diff
@@ -39,4 +36,64 @@
 </script>
 <!-- Intercom -->
 <script>(function(){var w=window;var ic=w.Intercom;if(typeof ic==="function"){ic('reattach_activator');ic('update',intercomSettings);}else{var d=document;var i=function(){i.c(arguments)};i.q=[];i.c=function(args){i.q.push(args)};w.Intercom=i;function l(){var s=d.createElement('script');s.type='text/javascript';s.async=true;s.src='https://widget.intercom.io/widget/rzcjacgd';var x=d.getElementsByTagName('script')[0];x.parentNode.insertBefore(s,x);}if(w.attachEvent){w.attachEvent('onload',l);}else{w.addEventListener('load',l,false);}}})()</script>
+<script>
+  (function ($) {
```
brikis98: Why move this from `main.js` and embed in HTML? Just to ensure it loads after Intercom code? If so, probably cleaner to define a function in `main.js` that has all the code below and to just call that function in the `script` tag here.

drmzio: Moved the cookie code to its own file so it can load asynchronously and won't affect load time.

_includes/scripts.html
drmzio: This cookie script was taken from the existing one that was found in main.js

_includes/scripts.html
drmzio: Same as above, I took the already existing cookie script and moved it to cookie.js

_includes/scripts.html
brikis98: Yes, I understand that, but boyscout rule says: leave it cleaner than you found it.

_includes/scripts.html
brikis98: Same here. It looks like a potential bug. Just because it was there before doesn't mean we should leave the bug in place!

_includes/scripts.html
sahilakos: @brikis98 The getCookiebyName function uses Reg Ex to grab all the cookies, and is later used to check if _GruntyCookie_ already exists or not. It was inspired by this [stack overflow answer](https://stackoverflow.com/questions/5639346/what-is-the-shortest-function-for-reading-a-cookie-by-name-in-javascript), but in short:

> regex approach is not only the fastest in most browsers, it yields the shortest function as well. 

I think I was initially using a different way to check for cookies, but you had suggested I go the RegEx route, so we had gone around this way. 

Here, @drmzio simply took the code it out from main.js and placed it inside scripts to load it at a later time for it to be initiated after intercom scripts, so the cookie notice can appear on top with a higher Z-index.

_includes/scripts.html
sahilakos: @brikis98 We're passing the name of the cookie we need to check for through the call below, so the first occurrence of the cookie name present validates our check to show or not show the modal.

_includes/scripts.html
brikis98: The StackOverflow link and the context you provided is exactly the kind of thing that makes for a great comment in the code.

_includes/scripts.html
brikis98: Not sure I follow. `match` will return `null` or an array. You're checking for `null` but not checking the array length before returning `pair[1]`. Yes, per the current regex design, the array _should_ have a length of 2, but that's the thing about bugs: it's possible that it won't, especially as this code is modified over time. Checking that `pair.length >= 2`, in addition to the fact that `pair` is not `null` seems worthwhile to avoid out of bounds errors, right?

sahilakos:
@brikis98 @josh-padnick Can you check this again. We have the printer-friendly branch too, but I'll update that and submit a different PR for it after I verify changes to upstream master versus our fork.

_includes/scripts.html
josh-padnick: @sahilakos Can you add the a comment here as Jim is requesting?

josh-padnick: Oh, actually looks like you added a comment. Nothing further needed here!

_includes/scripts.html
josh-padnick: @sahilakos This looks good now.

_includes/scripts.html
josh-padnick: Oh, are you saying that the `main.js` file may take a while to load, whereas the `cookie.js` file being shorter will load right away? Either way, we're already loading a bunch of separate javascript files (albeit mostly from third parties) so I think this is fine for now.

josh-padnick:
Ok, this looks good to me. I'll merge now and deploy!

PR #104, opened 2019-03-09
Add trivial formatting fixes to policy pages no one will read anyway.

josh-padnick:
OCD compels me to make these changes. Yes, it was worth the 10 minutes. Definitely. FYI, added these when previewing the Akos cookies changes. Now if someone actually clicks on the cookie policy, it will look like we care, which we do.

### DIFF
--- a/assets/css/app.less
+++ b/assets/css/app.less
@@ -14,4 +14,6 @@
 @import "global";
 
 // Import page-specific styles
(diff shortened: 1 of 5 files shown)


josh-padnick:
Merging this into my working branch, `gcp-tweaks`. Will be reviewable prior to merge to master.

PR #107, opened 2019-03-30
Move helm-kubernetes-services to open source module

yorinasub17:
I just realized `helm-kubernetes-services` is still labeled as a private beta repo, which is no longer true. So I moved it to the open source section.

That said, I am not sure if this is the best way to do this, since now it is not together with the EKS and Helm stuff. Let me know if you think this should be handled differently.

### DIFF
--- a/_data/library-infrastructure-packages.yml
+++ b/_data/library-infrastructure-packages.yml
@@ -67,16 +67,6 @@
     Deploy a best-practices Tiller (Helm Server) to your Kubernetes cluster. Supports namespaces, service accounts,
     least privilege RBAC roles, and automated TLS management.
 
-- name: "Kubernetes Services"
-  tags:
-    - K8S
-  private_beta: true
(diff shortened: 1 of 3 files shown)


josh-padnick:
In general, this page needs a redesign. :) But for this particular case, perhaps we can list the repo twice: once in the library and once in the open source section?

yorinasub17:
> But for this particular case, perhaps we can list the repo twice: once in the library and once in the open source section?

Done. It is a bit confusing still because you get the pop up, but also more visible this way. I removed the `private_beta` marker for helm kubernetes services. I plan on removing it for EKS as well once I am done integrating into the ref arch.

josh-padnick:
Roger. Did you deploy this to live yet?

josh-padnick:
Oh, I see you haven't merged. Ok, this LGTM to me at this point.

yorinasub17:
Ok merging and deploying. Thanks!

PR #110, opened 2019-04-11
Fix Docker Error

hngerebara:
Running ```docker-compose up``` without images previously built generates the error below
```Failed to fetch http://deb.debian.org/debian/dists/jessie-updates/InRelease  Unable to find expected entry 'main/bcted entry 'main/binary-amd64/Packages' in Release file (Wrong sources.list entry or malformed file```

Fixed using this answer [here](https://superuser.com/questions/1423486/issue-with-fetching-http-deb-debian-org-debian-dists-jessie-updates-inrelease) 

This issues also happens in ```houston-app```, therefore, I will create a similar PR once this is approved

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.4.2
+FROM ruby:2.6.2-stretch
 MAINTAINER Gruntwork <info@gruntwork.io>
 
 # Copy the Gemfile and Gemfile.lock into the image and run bundle install in a way that will be cached
@@ -9,8 +9,7 @@ RUN bundle install
 
 # Install Java, which is required for s3_website
 # Install texlive libraries, which are required for Pandoc
-RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list &&\
-    apt-get update && \
+RUN apt-get update && \
     apt-get install -y default-jre && \
     apt-get install -y texlive texlive-xetex texlive-latex-recommended texlive-latex-extra texlive-lang-cjk && \
     rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -9,7 +9,8 @@ RUN bundle install
 
 # Install Java, which is required for s3_website
 # Install texlive libraries, which are required for Pandoc
-RUN apt-get update && \
+RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list &&\
```
brikis98: Ah, yes, we've hit this issue with Jessie a few times! Sorry you ran into that.

I think the simpler fix is actually to update to a newer Debian version. In the `FROM` block at the type, I think you can switch to `ruby:2.6.2-stretch`.

Etiene:
Interesting... This issue was not supposed to happen on houston-app anymore. We were having that issue and Matt worked on a fix... It's unfortunate that it is resurfacing. Let me try to find some history of what went on

brikis98:
@Etiene This isn't the houston-app repo, it's the Gruntwork website.

Etiene:
@brikis98 From the PR description:
> This issues also happens in houston-app, therefore, I will create a similar PR once this is approved

brikis98:
Oh, my bad, totally missed that.

Dockerfile
hngerebara: I removed the fix I included and tried again using ```ruby:2.6.2-stretch``` but it doesn't seem to fix the issue.

brikis98: What is the issue you see with `-stretch`? Did you re-run `docker-compose build`?

hngerebara: It actually works. I had to remove the previous containers and images. I've updated the PR

brikis98:
BTW, feel free to merge your PRs once approved!

PR #113, opened 2019-04-16
Add profile for ore

oredavids:
#### What does this PR do?
- Adds Ore's profile to the About page
- Reduces the size of assets folder by ~840kb
   - Removes Toni's unused headshot image
   - Resizes headshot images for Hope and Rob

#### Screenshots
Screen Shot 2019-04-16 at 11 37 20 AM
Screen Shot 2019-04-16 at 11 37 00 AM


oredavids:
@brikis98 
- I meant to reduce Rob's image as well it was around 176kb, now down to ~22kb
- The gruntwork terms removal was unintentional, its been returned now.

brikis98:
I'm still seeing the terms as deleted in this PR?

oredavids:
@brikis98
It shows as a delete because a new terms file was generated after running `docker-compose up` when running locally. Should I have discarded the newly generated file?

brikis98:
Oh, I gotcha. Yea, if you didn't change anything in our terms of service, I'd exclude the Word Doc from the PR. It's probably some unimportant diff that Git can't see and ignore properly as it's a binary file.

josh-padnick:
We were trying to get clever and auto-generate a MS Word version of our terms of service very time the terms-of-service.yml file changes. We mostly got it working but it generates a new Word file _every_ time, regardless of whether there are changes in the yml file. So as Jim said, you can ignore the new Word file unless you’re actually making changes to the yml file and the updates to the Word file are meaningful

PR #117, opened 2019-05-09
Clarify Dedicated Support vs. Enterprise Support. Update to latest Terms of Service

josh-padnick:
This PR was initially intended to align what is today advertised as Dedicated Support with our internally discussed vision for Professional Support vs. Enterprise Support. In https://github.com/gruntwork-io/gruntwork-io.github.io/commit/e7d541e673840c5f59c36afe423cc59ac0ee5085, I made a simple HTML update to the support page to that effect.

But then I realized that the Terms of Service was also now out of date and needed to be updated. That, in turn, led me to realize that the previous Terms of Service updates I had discussed with Kevin were not incorporated on the site, and so I've now added those as well in https://github.com/gruntwork-io/gruntwork-io.github.io/commit/69904720762b12c56ea780bfc5d609a4e30c4bff, plus an updated MS Word file in 20f51b874ad249d69a74dc72be4e70a0ed5554c5.

These updates then made other aspects of our site outdated, and the following will need to be changed, either as part of this PR or a soon follow-up:

1. The Terms of Service has renamed "Dedicated Support" to "Professional Support." We should either update all public mentions of "Dedicated Support" to "Professional Support" or rename "Professional Support" to "Dedicated Support" in the Terms of Service.

2. Dedicated Support no longer includes phone/video, code reviews and other items. Various summaries of Dedicated Support on the site need to be updated.

3. The Terms of Service no longer include Houston with the Gruntwork Subscription. We should update the pricing and checkout pages to reflect this.

Because @robmorgan is actively updating our pricing and checkout pages, I've held off on these updates to avoid a merge conflict, but they should be incorporated as soon as possible.


josh-padnick:
In the interest of being a human-friendly company, I also added a CHANGELOG of all updates to our Terms of Service to make things more transparent:

image

I initially asked Akos to add this, and I cleaned up their work and filled in the right data.

josh-padnick:
This code will be reviewed again as part of an upcoming PR (#119?). Therefore, I'm merging it into my working branch now.

PR #121, opened 2019-05-16
More improvements to IaC Lib page

brikis98:
1. Track searches with Google Analytics events. Now we'll know what our website visitors are really looking for!
1. Debounce searches so we do searches and fire GA events no more than once every 250ms rather than every keystroke.
1. Add the list of submodules to each module so there is more detail on what we do and more content to search.
1. Rework UI to use "tags" and a cleaner table layout.

Here's what it looks like now:

Kapture 2019-05-16 at 1 28 39

This moves us one step closer to deprecating the TOC page. In fact, if we can put some of the instructions in it (e.g., how to consume our modules) into the docs site, we'd be pretty damn close...

### DIFF
--- a/_data/library.yml
+++ b/_data/library.yml
@@ -10,6 +10,25 @@
   description: |
     Create a best-practices Virtual Private Cloud (VPC) on AWS. Includes multiple subnet tiers, network ACLs, security
     groups, NAT gateways, Internet Gateways, and VPC peering.
+  submodules:
+    - name: vpc-app
+      blurb: launch an app VPC
+      description: Launch a VPC meant to house applications and production code. This module creates the VPC, 3 "tiers" of subnets (public, private app, private persistence) across all Availability Zones, route tables, routing rules, Internet gateways, and NAT gateways.
+    - name: vpc-mgmt
+      blurb: launch a mgmt VPC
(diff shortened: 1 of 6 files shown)


brikis98:
I'm going to merge and deploy this, as a customer asked for our list of modules, and I'd much rather send this than the TOC page. Feedback, however, is **very** welcome!

josh-padnick:
Whoah this looks great! Some quick UI-only feedback:

- When you click on one of the tags, can you auto-update the search bar?
- Can you make the AWS tag orange since that's their color? GCP could be red. Azure can be blue someday.
- Did you test with mobile?

josh-padnick:
- Also, when you press enter on your search the page scrolls to the very top, which is a bit jarring. 
- Typing "aws" appears to do nothing when in fact there are just more results than the viewport. Perhaps add "Showing X packages" so users can see that update in real-time?

brikis98:
> When you click on one of the tags, can you auto-update the search bar?

I wanted to, but it's a bit messy with the way the table rows themselves are clickable right now, so it'll take more time than I have. It's definitely a good idea though.

> Can you make the AWS tag orange since that's their color? GCP could be red. Azure can be blue someday.
> Did you test with mobile?

https://github.com/gruntwork-io/gruntwork-io.github.io/pull/122

brikis98:
> Also, when you press enter on your search the page scrolls to the very top, which is a bit jarring.
> Typing "aws" appears to do nothing when in fact there are just more results than the viewport. Perhaps add "Showing X packages" so users can see that update in real-time?

Also in https://github.com/gruntwork-io/gruntwork-io.github.io/pull/122

PR #122, opened 2019-05-16
A few follow-up tweaks to the IaC Lib based on PR feedback

brikis98:
Based on feedback in https://github.com/gruntwork-io/gruntwork-io.github.io/pull/121:

1. Change label colors so that AWS is orange, GCP is red, and Azure is blue.
1. Improve mobile rendering.
1. Show number of search results as you search.
1. Added a few more repos/submodules.

Kapture 2019-05-16 at 2 26 54


josh-padnick:
Looks great!

brikis98:
Also, just fixed the form behavior so hitting "enter" doesn't do anything weird.

brikis98:
Now deployed

PR #123, opened 2019-05-16
Fix javascript bug caused by #121

josh-padnick:
Following #121, every page on the site had this error in the web developer console:

image

The consequence of this bug was that subsequent javascript would fail to run, which disabled the ability to link to any section on the site.

The fix was to use the javascript specific to the IAC Library page on that page only.


josh-padnick:
Honestly, I get excited every time I catch a bug from you, because it just doesn't happen that often. :)

josh-padnick:
Merging and deploying...

PR #132, opened 2019-08-01
Remove houston listing

oredavids:
#### What does this PR do?
- Removes Houston listing from how-it-works and pricing pages


README.md
```diff
@@ -18,6 +18,8 @@ The fastest way to launch this site is to use [Docker](https://www.docker.com/).
 
 ## Manual quick start
 
+Recommended for use in development for faster bundle time and stylesheet reloads after changes
```
brikis98: I actually do not recommend we use the manual version. It will mess us up in terms of Ruby/Jekyll/Gem versions, missing configs, etc. 

What do you mean by "bundle time" here? And how much faster are the reloads? Have we tried any of the [docker perf improvements](https://github.com/gruntwork-io/mission-control/blob/master/docker-compose.yml#L11-L13)?

oredavids: Screen Shot 2019-08-01 at 12 04 17 PM
Screen Shot 2019-08-01 at 11 59 43 AM



@brikis98 It takes slightly over a minute to start up with docker and reloads can be anywhere from 9 to 50seconds depending on what else is running. The manual quickstart takes around 6seconds to start up and reloads are between 1.5 to 3 seconds.

I do however now see the issue with Gem versions so i'll remove that commit from this PR.

PR #133, opened 2019-08-01
Refactor docker-compose to make it much faster

brikis98:
@oredavids Brought up that using Jekyll + Docker was running very slowly for him: https://github.com/gruntwork-io/gruntwork-io.github.io/pull/132#discussion_r309643279

In local experiments, I was able to speed up my local dev speed by 5-10x: 14 second boot -> 2 second boot, 3 second hot reload -> 0.7 second hot reload.

1. Install the latest version of Docker: 2.1.0.0. This seemed to shave a couple seconds off Jekyll boot time.
1. Add `:delegated` to the bind mount. This also shaved a few seconds off boot time.
1. Use a docker-only volume for the generated `_site` folder. This shaved a few more seconds off the time.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,23 @@
-web:
-  build: .
-  volumes:
-    - .:/src
-  ports:
-    - "4000:4000"
-  environment:
-    - JEKYLL_ENV=development
-    # Pass your AWS credentials as env vars through to the Docker container for uploading the website to S3
-    - AWS_ACCESS_KEY_ID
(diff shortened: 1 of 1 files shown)


brikis98:
Also, the reason Docker is slow on Mac / Windows is due to poor file system performance for bind-mounting. If none of the fixes we do above help, you can also look into http://docker-sync.io/, which should give you near-native performance.

oredavids:
Much better.
Screen Shot 2019-08-01 at 12 44 57 PM

robmorgan:
Nice 👍

PR #135, opened 2019-08-02
Fix optimizely ad blocker issue

oredavids:
#### What does this PR do?
- Shows default content for scenarios where optimizely assets are blocked

gruntwork-default io


Etiene:
Great job 🎉

PR #144, opened 2019-08-20
Remove optimizely experiments

oredavids:
### What does this PR do?
- Removes optimizely experiments setup code
- Sets the beta nav and beta checkout flow as defaults


pages/checkout/index.html
```diff
@@ -1,15 +1,15 @@
 ---
 layout: default
-title: Become a Gruntwork Subscriber
-excerpt: Configure the plan just right for your team.
+title: Customize your plan with add-ons
+excerpt: We offer professional support and a reference architecture.
```
brikis98: Is `/checkout` the first page of the flow or the second now? This seems like the title for the second page, no?

oredavids: `/checkout` page is the second page of the flow now. `/pricing` is the first page. The title of the checkout page was changed in the new design. https://projects.invisionapp.com/share/EDSPCQC9QBX#/screens/370853456

oredavids:
@brikis98 A correctly structured checkout page url contains a query param like this:  `http://localhost:4000/checkout/?subscription-type=aws`. The query param is applied based on the user's cloud selection from the pricing page.

Effectively, a user directly accessing`/checkout-beta/` or now `/checkout/` would have a broken experience since they cannot pick a cloud provider on this page in the new design.

So first question; were the url links you shared with the customers correctly structured?

Also i tried jekyll's redirects and noted that if I crafted the checkout-beta url e.g `/checkout-beta/?subscription-type=aws` I end up being redirected to `/checkout/` - without the query params.

oredavids:
@brikis98 A new `gruntwork-terms.docx` file is generated for each run of `docker-compose up` I accidentally committed a newly generated one.

brikis98:
Ah, sorry, I linked customers to https://gruntwork.io/pricing-beta/, not checkout-beta.

pages/checkout/index.html
brikis98: Roger

brikis98:
So I guess we should add a redirect for `/pricing-beta` to `/pricing`. 

As a more general philosophy, think of URLs as a public API, and try to avoid making backwards incompatible changes (i.e., breaking URLs that used to work before).

oredavids:
@brikis98 this is ready for a final pass.

PR #148, opened 2019-08-23
Improve dev environment (Hack Day Project)

oredavids:
### What does this PR do?

- Fixes [Jekyll symlink security Vulnerability](https://github.com/gruntwork-io/gruntwork-io.github.io/network/alert/Gemfile.lock/jekyll/open)

- Fixes issue with [css changes not reflecting after file regeneration](https://jekyllrb.com/docs/configuration/incremental-regeneration/) unless the docker container is restarted. 

- Adds a browser reload after file changes and regeneration. 

#### Screencast Link
https://drive.google.com/drive/u/0/folders/1YWr46gSbZXidJ-VQWnrnO8mDlvXPSkLL


oredavids:
I think now may be a good time to have different docker-compose files so the command for dev can  be isolated.

_config.yml
```diff
@@ -1,5 +1,5 @@
 # Website settings
-baseurl: "/"
+baseurl: ""
```
brikis98: What does this change do?

jekyll-serve.sh
```diff
@@ -6,4 +6,4 @@ bundle exec jekyll build
 bundle exec pandoc -s --from html _site/terms/word.html --reference-doc=assets/msword/styles.docx -o assets/msword/gruntwork-terms.docx
 echo -e "\e[1;32mA new Terms of Service document (/assets/msword/gruntwork-terms.docx) has been generated."
 echo -e "\e[1;31mRun Jekyll serve to watch for changes"
-bundle exec jekyll serve --drafts --incremental --host 0.0.0.0
+bundle exec jekyll serve --livereload --drafts --host 0.0.0.0
```
brikis98: Why not `--incremental`? Is that what breaks CSS reload?

_config.yml
oredavids: From version 3.3, it is recommended in the jekyll docs, to never set the baseurl to "/". I discovered the issue when trying to get the browser reload working. The server address ended up with a double slash at the end like this: `http://0.0.0.0:4000//` when the baseurl is set to "/".

https://jekyllrb.com/news/#2-relative_url-and-absolute_url-filters

jekyll-serve.sh
oredavids: Yes, this is what broke the CSS reload. The `--incremental` flag is still experimental and only watches dependencies for `includes` and `layout` files.

https://jekyllrb.com/docs/configuration/incremental-regeneration/

robmorgan:
Nice 👍

jekyll-serve.sh
brikis98: Gotcha. What impact does this have on hot reload performance?

_config.yml
brikis98: Understood, thx!

jekyll-serve.sh
oredavids: It increases hot reload by approximately 1 second.

brikis98: Hm, OK... We should keep our eye on this. As the amount of content goes up (which it will with the deployment guides!), this overhead will get worse without `--incremental`.

PR #153, opened 2019-09-09
Modify table styling

hngerebara:
Fix table styling to match [mockup](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d6599b8c463d49b78307757)
Screenshot below:

Screenshot 2019-09-09 at 8 08 31 AM

N/B: I remove unused code for deploy cta and formatted the code

### DIFF
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1476,9 +1476,6 @@ body .guides .guides-section-white {
 
 .post-bg{
   background-color: #f6f8fa; 
-  .container-fluid {
-    max-width: 1200px;
-  }
 }
 
 .badge {
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1722,7 +1722,7 @@ body .guides .guides-section-white {
   }
   .imageblock > .title {
       font-size: 12px;
-      margin: 10px 0;
+      margin: 5px 0 15px 0;
       color: rgba(0, 0, 0, 0.54);
       text-align: center;
   } 
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1720,6 +1720,11 @@ body .guides .guides-section-white {
       letter-spacing: 0.39px;
     }
   }
+  .imageblock > .title {
+      font-size: 12px;
+      margin-bottom: 10px;
+      font-weight: 600;
+  } 
   .breadcrumb {
     font-size: 14px;
     font-weight: bold;
@@ -1760,6 +1765,7 @@ body .guides .guides-section-white {
   }
 }
 
+
 // Post Detail styles
 .post-content {
   padding: 0 2em;
@@ -1776,10 +1782,11 @@ body .guides .guides-section-white {
   }
   img {
     max-width: 100%;
-    padding: 15px;
+    padding-top: 10px;
     display: block;
     margin: 0 auto;
   }
+ 
   pre, pre[class*="language-"] {
     padding: 0px;
     background: aliceblue;
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1722,8 +1722,9 @@ body .guides .guides-section-white {
   }
   .imageblock > .title {
       font-size: 12px;
-      margin-bottom: 10px;
-      font-weight: 600;
+      margin: 10px 0;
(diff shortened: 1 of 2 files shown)


brikis98:
Thanks Hope! A couple thoughts:

1. The column headings should only have the blue underline when we have a special "hover" behavior to show more info. We have such a behavior on the detailed comparison page (that you're using as the inspiration for this styling), but not on deployment guides.
1. The vertical spacing between rows made sense on the comparison page where we have the logos and names for each row in the left-most column (e.g., ECS, EKS, etc). Without that, the rows feel too far apart. Could you reduce the vertical spacing by about half?

hngerebara:
The hover effect displays a single line not dashed. Should that be the case here as well? @brikis98

brikis98:
I'm saying there should be no hover effect at all. It's only there when we have more information we want to show the user about that column. We don't have any such information for the deployment guides, so no underline, dashed or solid, and no hover effect on these pages.

hngerebara:
Reduced the font for image title too. Screenshot below

Screenshot 2019-09-09 at 10 10 10 AM

brikis98:
> Reduced the font for image title too. Screenshot below

Nice. I wonder if we shouldn't copy Medium's style here for image captions: that is, a small, light gray font, with some margin below it to separate the text from the image. See https://blog.gruntwork.io/how-to-manage-terraform-state-28f5697e68fa for an example.

hngerebara:
Done: See below

Screenshot 2019-09-09 at 10 32 36 AM
Screenshot 2019-09-09 at 10 36 44 AM

brikis98:
Nice! Make sure the caption is closer to the image than it is to the surrounding text. So `margin-top` should be fairly small (`5px` perhaps?) and `margin-bottom` should be larger (`10px` perhaps?).

Etiene:
Once this is done, should we ask Sam to update the designs on zeplin to reflect the decision made here? I agree with the small gray subtitle for the images, but if we are going to diverge from the design are we at risk from accidental regressions in the future?

hngerebara:
There was no design for image titles. I guess we can wait till she gets back to confirm.

assets/css/global.less
```diff
@@ -1401,6 +1401,9 @@ body .guides .guides-section-white {
 
 .post-bg{
   background-color: #f6f8fa; 
+  .container-fluid {
+    max-width: 1200px;
```
Etiene: What's the intention of this line? It looks like this is causing an issue with the post width (it shrunk considerably on my screen)

Etiene: Before this commit:
Screen Shot 2019-09-09 at 18 16 05

After this commit:
Screen Shot 2019-09-09 at 18 22 04

hngerebara: I wanted to maintain a max-width for both detail and listings page so it doesn't stretch so much on larger devices. The listings page also has a max-width of 1200px.

Etiene: My device is not large though, I'm on a 13' macbook, before my screen was matching the mock and now it is not anymore. Since this is causing some screens to be buggy and this is not related to "table styling" anyway, I suggest you to remove this from this PR, and we discuss this separately. I also think we should get a browserstack subscription and see how it works on our screens!

brikis98:
> Once this is done, should we ask Sam to update the designs on zeplin to reflect the decision made here? I agree with the small gray subtitle for the images, but if we are going to diverge from the design are we at risk from accidental regressions in the future?

Yup, that's a good idea

assets/css/global.less
brikis98: We do need a max width, as on my 1920x1080 screen, it ends up super wide, but +1 on doing that in a separate PR.

hngerebara: Task created here to fix this https://gruntwork.atlassian.net/browse/HSTN-317

hngerebara:
Created a ticket to get Sam's opinion about the table title styling here https://gruntwork.atlassian.net/browse/HSTN-365

assets/css/global.less
Etiene: Awesome! Once this is removed from this PR then I think this PR is good to go then. Or is there anything else pending?

PR #147, opened 2019-08-23
implement searching

hngerebara:
Things done in this PR

Design to match 
[Listings view](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283d0b6b443505e2853d) 

[Detail view](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283e91c8889afa13f5f2)

[Newsletter modal view](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283c8d5c26520ba7472a)

[No results view](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283dccabfe9c037443a4)

Only the Filter dropdown [here](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d65212c9df85c9b3ca072f8)

[Table view](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d6599b8c463d49b78307757)
Implemented search and filtering by tags and cloud

To be completed today:
Google analytics

 https://92eca3a4.ngrok.io

PR #155, opened 2019-09-09
Fix callouts

hngerebara:
Done the following in this PR

- Breadcrumb underline should not include the slash

Style breadcrumb as given in mockup
(Note: Using ```font-weight: bold``` when hover or active made the bread crumb jump. So I used ``shadow: 0.5px``` instead)

- Match title in the breadcrumb to post title

- Fix callouts to display icon not text

- Style callout to match mockup

Screenshots below:

Breadcrumb:
Screenshot 2019-09-09 at 2 50 18 PM

Callout
Screenshot 2019-09-09 at 1 39 14 PM

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -19,9 +19,9 @@ custom_js:
         <a href="/guides" ga-event-category="guide-{{ post.title | downcase}}" ga-event-action="deployment-guides">Production Deployment Guides</a> 
         {% for crumb in crumbs offset: 2 %}
           {% if forloop.last %}
-           <a href="#"> / <span> {{ crumb | replace:'-',' ' | capitalize }}</span></a>
+           <a href="#"> / <span> {{ page.title | replace:'-',' ' }}</span></a>
           {% else %}
-            <a href="/guides#{{ crumb }}" >/ <span class="active">{{ crumb | replace:'-',' ' | capitalize }}</span></a>
+            <a href="/guides#{{ crumb }}" >/ <span>{{ crumb | replace:'-',' ' | capitalize }}</span></a>
           {% endif %}
         {% endfor %}      
       </ol>
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -14,17 +14,17 @@ custom_js:
       </div>
     </div>
     <div class="col-md-7 guides-section-white">
-      <ol class="breadcrumb text-center">
-        {% assign crumbs = page.url | split: '/' %}
-        <a href="/guides" ga-event-category="guide-{{ post.title | downcase}}" ga-event-action="deployment-guides">Production Deployment Guides</a> 
-        {% for crumb in crumbs offset: 2 %}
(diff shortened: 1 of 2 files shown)


Etiene:
Fix for breadcrumb looks good! Good catch on the jumpiness 👍 

For the callouts, fix for the "!" symbol looks good! I have some other requests:

1. Could you please also fix the margin around the callout?

If you look at the design on [Zeplin](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283e91c8889afa13f5f2) it will show you the exact distance of the mock if you click on the element then hover the mouse around other things:

Screen Shot 2019-09-09 at 17 35 39


2. Could you make one issue per PR? For example, in this case, if these were two PRs, I could have approved the breadcrumb one already

And some questions:

1. Do you know why only some callouts have the correct font? I found this one that has font-size 16px instead of 12. It's on how-to-deploy-production-grade-vpc-aws

Screen Shot 2019-09-09 at 17 32 46

2. On how-to-use-gruntwork-service-catalog the callouts in a list are following the list indentation. This is not a detail that the mocks answer and it doesnt have any example for this particular case. But I wonder if this is what we should be doing in this case... Maybe it should live "outside" of the list and have the full length like the other callouts. What do you think? Should we ask Sam?
Screen Shot 2019-09-09 at 17 40 46

hngerebara:
@Etiene Comments for your request
1. If I implement the exact margin as given in the mockup, we will have something like the picture below for callouts that are separated by a sentence.

Screenshot 2019-09-10 at 5 48 49 AM

2. I'll create a new PR for breadcrumbs

Answers to the questions above

1.  Thanks for spotting that. Seems that callout was picking the parent font-size of 16px while others were actually using the exact font-size assigned. I will properly target the callout paragraph 

2. Personally I feel if it is a list, it means the callout should be within that list-item. I will have a conversation with Sam tomorrow when she returns because Jim also felt we shouldn't have list items indented https://github.com/gruntwork-io/gruntwork-io.github.io/pull/147#issuecomment-528312973

Also using the exact font-weight as given in the mockup, makes the text too bold. Is this how we want the callouts? Screenshot above.

Etiene:
@hngerebara 
About the requests
1. Yes, I think that looks ok

2. Oh, you didnt have to separate this one already, just the future ones!  My bad, I think I didnt make this clear. Thanks!

questions

1. Cool! I'll have a second look

2. Ok, let's wait on that one.


About your question:

> Also using the exact font-weight as given in the mockup, makes the text too bold. 

I agree with that! Leave the previous font-weight as visually it looks more similar to the mockup, but later we should maybe investigate why there is this difference (I dont think it is worth the time right now)

hngerebara:
Created questions for Sam to address in these tickets
https://gruntwork.atlassian.net/browse/HSTN-366
https://gruntwork.atlassian.net/browse/HSTN-364

PR #158, opened 2019-09-11
resize image and modify listing styling

hngerebara:
Resized thumbnails to be in the middle.
Screenshot 2019-09-11 at 5 16 01 AM


Etiene:
We should probably standardize the size of these images that we use for thumbnails (at least the height x width proportion) so they don't stretch weirdly like this amazon one... We could try to find a way to achieve this through css, but we still need to resize them anyway (some of these are very big images). Could you take care of that as part of this PR?

assets/css/global.less
```diff
@@ -1227,14 +1229,37 @@ body .guides .guides-section-white {
...
+      @media (max-width: @screen-xs-min) {
+        .remove-visibility();
+      }
       max-height: 100px;
+      max-width: 100%;
+      max-height: 100%;
+      display: block;
+      margin: auto auto;
+      padding: 15px;
+      height: 85px;
```
Etiene: Max-height is duplicated here, and if you are going for a fixed height, it both max and height and width are now unnecessary. Also the height might need to be a bit bigger... Something like 1-3px more? Still seems like it's slightly off, but I would get the final thumbnail images first before working on this

hngerebara:
Yes done it already in css. Currently, all images have the same height but because some of the original images uploaded are rectangles, giving a fixed with will make it look a lot weird. I can manually fix this now by making the image a square and assigning a fixed with but subsequent pictures to be uploaded need to be squared so they don't look weird

Etiene:
Let's punt on this one for now, and let's try to work this out after I'm done with the ref arch?

hngerebara:
@Etiene Any updates here? Should I go ahead with what I suggested earlier?

brikis98:
Sanity check: did you look at https://getbootstrap.com/docs/4.0/content/images/?

Etiene:
@brikis98 not yet! I picked this from hope while she works on the category scrolling and I started working on it yesterday. I was close to being satisfied with the results, but I'll see if it can get even better by applying some of those classes. 👍

Etiene:
## Before
Screen Shot 2019-09-17 at 12 30 09
Screen Shot 2019-09-17 at 12 14 46

## After
Screen Shot 2019-09-17 at 12 28 13
Screen Shot 2019-09-17 at 12 26 02

Etiene:
Had to rebase it again because #174 also introduced a conflict. I'll merge this soon if no-one objects and any other comments I can address later.

PR #159, opened 2019-09-11
Cloud filtering

hngerebara:
What this PR covers 

- Deselecting a cloud filter button to display all items that existed before filtering

- Hide filter results count if no results found

- Properly align search and filter to the middle to match the mockup

Video of how it works below:
https://drive.google.com/file/d/1xch0iejGTp7hNIC8crSr5KPjuf9JC4Lb/view?usp=sharing

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -218,6 +218,8 @@
 
     if ($(this).hasClass('initialSelect') && $(this).hasClass('active-button') ) {
       $(this).removeClass('initialSelect');
+      $(this).removeClass('active-button');
+      $('#no-matches').hide();
       showAllItems();
     } else {
       $(this).addClass('active-button');
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -26,6 +26,7 @@
   }
 
   function showItemsCount(totalCount, numSubmodules) {
+    $('#search-results-count').show();
     if (searchEntry.type == 'libraryEntries') {
       if (totalCount > 0 && numSubmodules > 0) {
         $('#search-results-count').html("<strong>" + totalCount + "</strong> repos (~<strong>" + numSubmodules + "</strong> modules)");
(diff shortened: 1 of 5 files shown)


Etiene:
Yay! 
>Deselecting a cloud filter button to display all items that existed before filtering
> Hide filter results count if no results found
> Properly align search and filter to the middle to match the mockup
✅ ✅ ✅ 

And the filters disappearing on no results when you filter by Azure bug is gone too! ✅ 

But I see another issue, the content in the "no results" gets jumpy when you scroll. Could you have a look at that? I can show you what I'm talking about in a call if you can't reproduce this. 

Question, are the guides supposed to befiltered by aws when you load the page? 
Because current it is not but the AWS button shows as selected.

Because here is one issue which is not in the scope of this card... In the way the filter is currently implemented now, our category separations disappear, we get all results together the "2 results found" in the corner. So filtered can't be our current default unless we refactor this.

I know @brikis98 has said in a previous PR that:
>There's no cloud filter selected by default... And once you select one, there's no way to deselect it. We probably want to have AWS selected by default (which I think is in the InVision design).

But 1. I'm not sure if that's what the invision design really is defines (it might just be showing how that filter button should look like when selected) and 2. the most important thing here is fixing the behavior where we were unable to deselect things, which was a bug that needed to be fixed in order to ship this, and which you have resolved in this PR. 

We have two options:

1. Remove the active-button class from the aws cloud filter and remove the behavior that reapplies this class to AWS once you unselect another button, which will actually make that code cleaner, and closer to getting this PR through. Then check with Jim and Sam later if the intention is really to filter on page load and create a separate card for that and work on that specifically on a future PR.
2. Check with Jim and Sam asap and refactor the search results on this PR to still show the categories and filter and apply the filter at page load.

assets/js/search.js
```diff
@@ -211,25 +210,31 @@
...
-    else if(id !== 'aws' && $(this).hasClass('active-button')){
-      $('.cloud-filter #aws').addClass('active-button');
-      $(this).removeClass('active-button');
+  function performSearch(filterButton) {
+    const id = filterButton.attr('id');
+
+    if (filterButton.hasClass('initialSelect') && filterButton.hasClass('active-button') ) {
+      filterButton.removeClass('initialSelect');
+      filterButton.removeClass('active-button');
+      $('#no-matches').hide();
```
Etiene: this code looks so much better now 👍

PR #165, opened 2019-09-13
Azure coming soon page

hngerebara:
Implement coming soon page for azure based on mockup [here](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283d82415f9b2a1f20b2)

https://drive.google.com/file/d/1FqwMfzVLrkR04VcWnj9bWszLGCwoOQMX/view?usp=sharing

### DIFF
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1460,6 +1460,7 @@ body .guides .guides-section-white {
   color: @guides-font;
   padding-bottom: 20px;
   font-family: @guides-font-family;
+  padding-bottom: 115px;
   .container-fluid {
     max-width: 1200px;
   }
@@ -1577,17 +1578,6 @@ body .guides .guides-section-white {
       background-image: linear-gradient(to bottom, #5bcde7, #569cea);
     }
   }
-}
-
-.overlay {
-  width: 100%;
-  height: 100%;
-  top: 0;
-  left: 0;
-  right: 0;
-  bottom: 0;
-  opacity: 0.2;
-  pointer-events: none;
 }
 
 .admonitionblock.important, .admonitionblock.note{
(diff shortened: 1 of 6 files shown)


Etiene:
1. The white overlay on guides needs to have a gradient that gets stronger on the bottom, until the posts disappear, it is not uniform
Screen Shot 2019-09-13 at 12 28 37

2. BUT, I have a question about this. Isn't it weird when you're switching from a different filter and have that content in the back change? I mean, nothing guarantees that the user is always coming from an unfiltered state into the azure filter. 

3. The tag filters stopped working :(

Etiene:
Background looks good now! 👍 

The tags still have a bug, when you deselect them, the results don't go back to what they were before

hngerebara:
Thanks. The ticket to fix it already exists here https://gruntwork.atlassian.net/jira/software/projects/HSTN/boards/31?selectedIssue=HSTN-362

assets/css/global.less
```diff
@@ -1544,6 +1545,41 @@ body .guides .guides-section-white {
...
+      display: none;
+    }
+  }
+  p {
+    font-size: 20px;
+    opacity: 0.7;
+    line-height: 1.5;
+    letter-spacing: 0.56px;
+    color: #000000;
+    margin-bottom: 24px;
```
Etiene: Does this part of the css still need to exist?

hngerebara: Yes it does

hngerebara: Screenshot 2019-09-17 at 4 22 28 PM

Etiene: Oh ok! I was confused because of the opacity, I thought this was about that old background

Etiene:
Ugh! I merged to the wrong branch. Will fix this in a follow up PR

PR #166, opened 2019-09-15
Subheadings to have anchor links

hngerebara:
Add anchor links to sub headings

Before:
Screenshot 2019-09-15 at 2 15 00 AM

After:
Screenshot 2019-09-15 at 2 14 48 AM


Etiene:
Thanks for the fix!
Some questions:

1. Is changing colors back to black on focus intentional?
2. It looks slightly misaligned at the bottom. Is that also intentional? Experimenting on it, giving it a 0.9em font size made it look more aligned. What do you think?

hngerebara:
1. According to Sam, the subheading test should not have any color on hover.
2. I'm not sure I understand this.
> It looks slightly misaligned at the bottom. Is that also intentional? Experimenting on it, giving it a 0.9em font size made it look more aligned. What do you think?

 What are you looking at?

Etiene:
> 1. According to Sam, the subheading test should not have any color on hover.

I didnt mean the heading, I mean the icon for the anchor link. It gets blue on hover, but goes back black on focus, this gives a weird flickering effect. I think this focus is inherited from somewhere else, so it might be a little hard to find.

> 2. I'm not sure I understand this. What are you looking at?

The screenshot you just posted and the branch on my local dev. The anchor link icon is slighly too low
64915439-baa84b80-d75e-11e9-976b-78c5d6c550b3

hngerebara:
FIxed this. You can view [here](https://drive.google.com/file/d/1d6jvsdu7nF2MNRqjTH-IPTb3yA6YxA38/view?usp=sharing)

assets/css/global.less
```diff
@@ -1970,6 +1970,13 @@ body .guides .guides-section-white {
   }
   h1, h2, h3, h4, h5, h6 {
     color: @guides-font;
+    margin: 10px 0;
```
brikis98: Why add a margin on top and bottom?

assets/css/global.less
```diff
@@ -1970,6 +1970,13 @@ body .guides .guides-section-white {
   }
   h1, h2, h3, h4, h5, h6 {
     color: @guides-font;
+    margin: 10px 0; 
+    &:hover > a {
```
brikis98: Where does the `a` come from? Is that added automatically by JS?

Etiene:
The alignment looks good on your video, but it's still off on my screen when I check out the branch. Are there any changes you didn't push?

assets/css/global.less
hngerebara: The mockup has some margin for all header tags.

assets/css/global.less
hngerebara: Jekyll adds an anchor tag to all headers

hngerebara:
@Etiene I still cannot replicate this issue and this branch is updated. Maybe we should quickly have a sync

assets/css/global.less
brikis98: Yes, but I believe the spacing for headers looks correct already in master, after a bunch of tweaks from yesterday. Can you show what change this is making?

PR #170, opened 2019-09-17
fix deselecting tags

hngerebara:
Enable tag filter deselection

https://drive.google.com/file/d/1OIrBw6DmGBbBBScMfHQljMDoxXtC0Hxq/view?usp=sharing


Etiene:
Found a bug, it reapplies the category filter without any signal that it did so.

Here's how to reproduce:

1 - load the page (it shows 2 posts in AWS)
2 - deselect aws cloud to see all posts (4 posts)
3 - click on filter and select networking (it will show only the networking post (1 post))
4 - deselect it (it shows 2 posts, but should show all)

Etiene:
Another bug:

1 - load the page (it shows 2 AWS posts)
2 - select GKE filter (it shows one of the GCP posts, but AWS cloud still shows as selected even though the GKE post is not an AWS post) <- Not sure what would be the best behavior here, there isn't a clear answer like the other post.

There are two options:
* Either we keep AWS cloud selected and show "No Results"
or
* It deselects AWS and shows the GKE post

Etiene:
FYI I will work on this one

Etiene:
Status: 
First bug fixed.
Second bug: discussing what solution. https://gruntwork-io.slack.com/archives/C76UJFUEL/p1568742841031900

Eugene proposed a third option: 
Have the cloud filter modify the pool of tags and not be able to select GKE at all if you have AWS selected

_posts/2019-05-08-deploying-a-dockerized-app-on-gcp-gke.md
```diff
@@ -4,7 +4,7 @@ type: guide
 title: Deploying a Dockerized app on GCP and GKE
 image: /assets/img/guides/gke.jpg
 excerpt: Before we can deploy a dockerized app, we need to first create one. For the purposes of this guide we will create a basic Node.js app that responds to requests on port 8080.
-tags: ["gcp", "gke", "docker"]
+tags: ["gcp", "gke", "docker", "kubernetes"]
```
brikis98: This will conflict with https://github.com/gruntwork-io/gruntwork-io.github.io/pull/182. So feel free to revert, and I'll add your fix to my branch.

assets/js/search.js
```diff
@@ -28,7 +28,7 @@
   /**
    * Function displays the total items displayed on the page
    */
-  function showItemsCount(totalCount, numSubmodules) {
+  function showItemsCount(searchEntry, totalCount, numSubmodules) {
```
brikis98: `searchEntry` isn't used here?

assets/js/search.js
```diff
@@ -37,7 +37,7 @@
         $('#search-results-count').text("0 repos");
       }
     } else {
-      if (totalCount > 0 && numSubmodules === 0) {
+      if (totalCount > 0 ) {
```
brikis98: Does this still work properly on IaC Lib page?

assets/js/search.js
```diff
@@ -55,25 +55,18 @@
...
 
-  let searchEntry;
-
   function initialEntry() {
     $('#no-matches').hide();
     $('#no-azure-results').hide();
-    searchEntry = detectSearchEntry();
-    $(showInitialItemsCount);
-    $(performSearch($('.cloud-filter #aws')));
+    selectCloud($('.cloud-filter #aws'));
```
brikis98: Add comment to make it clear we're intentionally selecting AWS as the default

assets/js/search.js
```diff
@@ -164,7 +153,7 @@
               searchContent = entry.title + entry.category + entry.content + entry.tags;
               break;
             case searchEntry.type === 'guideEntries' && type === 'tagSearch':
```
brikis98: Not related to this PR, but do we have a Jira card to clean this up? I mentioned a couple times to @hngerebara that this super-hardcoded stuff means this search code is neither generic/reusable, nor particularly easy to read/maintain. Happy to chat about how to improve.

assets/js/search.js
Etiene: this doesn't affect the iac lib page (if you expand the code above, you'll see this is in the `else` block, so numsubmodules is not relevant anymore, which is why I removed it

assets/js/search.js
Etiene: What are you talking about? It's in the line just below this one

brikis98: Oh... So how did this work at all before? I saw this variable added and figured it was new and was expecting usage of it to show up in the diff too... Was the code just throwing errors all the time instead?

assets/js/search.js
Etiene: I did some very small improvements on the parts that I touched or little things that would not alter the logic (such as this replacement of `.map` for `.forEach`), as I was fixed on fixing things asap. But yes, a better refactor would be ideal. Another needed refactor is splitting `global.less` into individual smaller files like we do with the js. I think hope created a card for that one, but I'm gonna ask her if there's a card for refactoring this too. Although to be completely honest I'm not sure if it's worth a lot of effort soon given we are migrating to a dynamic app and we can just make better choices and use an external lib or just plain react state management for things like this. I guess we should make some common sense approach depending how far we get from migrating to the dynamic app.

assets/js/search.js
Etiene: It was working because `searchEntry` was a global var set at the inital page load. I removed the global var and placed it under beter scoping.

assets/js/search.js
brikis98: +1 on figuring out if it makes sense to touch this further. Definitely not a high priority.

assets/js/search.js
brikis98: Ohhh, understood, thx

PR #172, opened 2019-09-17
Change to all rights reserved license

brikis98:
See https://gruntwork-io.slack.com/archives/C656LE2GG/p1568719756012700 for discussion.


Etiene:
I have some questions about the approach, but I'll bring those on slack

brikis98:
Note: checking with lawyers on this too.

brikis98:
I'm going to merge this for now, as I'm guessing it's better than what we had before, even if not perfect. I'll put updates in a future PR, if necessary, once I hear back from legal.

PR #178, opened 2019-09-18
Increase span callout font size

hngerebara:
Increase span callout font size

### DIFF
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1566,6 +1566,7 @@ body .guides .guides-section-white {
           color: @guides-font;
           font-family: AndaleMono;
           font-weight: normal;
+          font-size: 14px;
         }
         &:nth-child(even) {
           background-color: inherit !important;
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1566,7 +1566,7 @@ body .guides .guides-section-white {
           color: @guides-font;
           font-family: AndaleMono;
           font-weight: normal;
-          font-size: 14px;
+          font-size: 12px;
         }
         &:nth-child(even) {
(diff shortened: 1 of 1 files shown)


brikis98:
Could you share a before/after screenshot?

hngerebara:
**Before:** 
Screenshot 2019-09-19 at 10 30 17 AM

**After**
Screenshot 2019-09-19 at 10 30 01 AM

brikis98:
Looks a bit odd to have the code text bigger than the rest. Perhaps increase the font size for all the text in call outs to be the same (i.e., slightly larger than it was originally)?

Etiene:
The before one looks correct and better than the 14px one. But I clearly remember Jim finding one instance of this where the span was smaller than the rest. Was this something that's working differently on some screens? Was the problem one particular span? 

> Perhaps increase the font size for all the text in call outs to be the same (i.e., slightly larger than it was originally)? 

I don't think this is a solution to the original problem

brikis98:
> I don't think this is a solution to the original problem

Why not? The font is pretty small for callouts, so making it bigger doesn't seem to hurt... And would also make the font bigger for code spans within callouts.

Etiene:
Because the problem is here
Screen Shot 2019-09-19 at 11 03 11

brikis98:
Oh, yea, that's even better to fix 👍

hngerebara:
It still doesn't make any difference. Unless I'll change the entire font

hngerebara:
This has now been fixed. Had a sync with @Etiene to be certain
Screenshot 2019-09-19 at 12 45 24 PM

PR #179, opened 2019-09-18
Final edit pass for service catalog guide; various styling fixes

brikis98:
Did one last edit pass on the "How to use the Gruntwork Service Catalog" guide. As I edited it, I noticed various styling issues, and fixed those too:

1. Make titles in call-outs bold and fix alignment.
1. Increase spacing for lists.
1. Update our usage of prism.js to pull in only a few languages, rather than all. This reduces file size from ~350kb to ~25kb.
1. Switch to a different prism.js theme and override a couple settings to better fit the surrounding page.
1. Make headings bigger and increase spacing.
1. Increase line-height on all paragraphs to make reading more pleasant.

Headings before:

Screen Shot 2019-09-19 at 12 12 25 AM

Headings after:

Screen Shot 2019-09-19 at 12 11 11 AM

Code before:

Screen Shot 2019-09-19 at 12 12 16 AM

Code after:

Screen Shot 2019-09-19 at 12 11 56 AM


brikis98:
OK, going to merge this in so I can do the final edit passes on the remaining content. Feedback is still very welcome and will be addressed in follow-up PRs.

Etiene:
> OK, going to merge this in so I can do the final edit passes on the remaining content. Feedback is still very welcome and will be addressed in follow-up PRs.

Reviewing now. But you can also base PRs on open PRs.

Etiene:
Found an unstyled block at http://localhost:4000/guides/foundations/how-to-use-gruntwork-service-catalog

(comment should be in a different color?)
Screen Shot 2019-09-19 at 10 48 35

Etiene:
Just noticed this, related to a different PR, but since you're doing a pass on final styles I'm gonna post it here too (this one is affected by screen resolution)

http://localhost:4000/guides/foundations/how-to-configure-production-grade-aws-account-structure#create-iam-groups-iam-users-and-an-iam-password-policy-in-the-root-account

Screen Shot 2019-09-19 at 10 51 41

brikis98:
> Found an unstyled block at http://localhost:4000/guides/foundations/how-to-use-gruntwork-service-catalog

Good catch. Fixed in https://github.com/gruntwork-io/gruntwork-io.github.io/pull/181.

> Just noticed this, related to a different PR, but since you're doing a pass on final styles I'm gonna post it here too (this one is affected by screen resolution)

Ah, you're right. Filed https://gruntwork.atlassian.net/browse/HSTN-419.

PR #180, opened 2019-09-19
Use replaceState instead of pushState

brikis98:
This way, as you scroll, it doesn't go into your browser's history, and make the "back" button useless.


brikis98:
Thx for catching this and the review!

PR #182, opened 2019-09-19
Convert GCP post from Markdown to AsciiDoc

brikis98:
1. Convert from Markdown to AsciiDoc, as Jekyll produces different HTML for each one, and we've so far only tested the styling for the AsciiDoc flavor.
1. Convert the post to more closely match the format of other deployment guides. There's still a lot more we could do here, but that will have to be after launch.


robmorgan:
I'm not an expert on Asciidoc, but LGTM to me on a first pass.

brikis98:
Thx for the review!

PR #184, opened 2019-09-19
fix guide toc width for multiple screen sizes

Etiene:
Fixes the width of the toc

### DIFF
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -1175,6 +1175,10 @@ body .guides .guides-section-white {
 }
 
 #toc {
+  &.fixed {
+    width: inherit;
+    padding-right: 16px !important;
+  }
   .guides-outline;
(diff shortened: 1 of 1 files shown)


assets/css/global.less
```diff
@@ -1175,14 +1175,17 @@ body .guides .guides-section-white {
 }
 
 #toc {
+  &.fixed {
```
brikis98: Why does this only apply when it's in `.fixed` mode? That class is only added while scrolling... Not when at the top or bottom of the page.

Etiene: That's exactly when it messes up. When the fixed class is added it starts ignoring the parent width for some reason.

brikis98: Ugh... CSS is hard. OK, ship it! 🎉

PR #186, opened 2019-09-19
fix accordion error in console

hngerebara:
This PR fixes the accordion error in the console.

Some pages have a `.panel` class which doesn't exist in the deployment guides so I put a condition to check if guides is found in the path.

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -387,30 +387,31 @@ $(function(){
 });
 // open accordion for direct links to panel id
 $(window).load(function() {
-  var getHash = location.hash; //get hash from js object location
-  var accordion = $('a[href="' + getHash + '"]');
-  $panelHeight = accordion.parents('.panel').height();
+  const getHash = location.hash; //get hash from js object location
 
-  //Handle cases that do not return a null height
-  if($panelHeight){
-    function activateAccordion(id) {
-      if (id.length)//check if hash isn't empty
-      {
-        $('html,body').animate({
-            scrollTop: accordion.parents('.panel').offset().top //scroll to accordion
-        }, 100, function () {
-            accordion.click(); //simulate click
-            accordion.parents('.panel').find('.panel-title a.no-scroll i.fa').addClass('panel-isOpen'); //point caret down for oepned panel
-        });
-      }
-      else{
-        //if no hash is called, open first panel
-        $('.panel-group .panel:first-child .panel-collapse').addClass('in');
-        $('.panel-group .panel:first-child .panel-title a.no-scroll i.fa').addClass('panel-isOpen');
-      }
-    };
+  function activateAccordion(id) {
+    const accordion = $('a[href="' + id + '"]');
+    const $panelHeight = accordion.parents('.panel').height();
+
+    //Handle cases that do not return a null height
+    if($panelHeight){
+        if (id.length)//check if hash isn't empty
+        {
+          $('html,body').animate({
+              scrollTop: accordion.parents('.panel').offset().top //scroll to accordion
+          }, 100, function () {
+              accordion.click(); //simulate click
+              accordion.parents('.panel').find('.panel-title a.no-scroll i.fa').addClass('panel-isOpen'); //point caret down for oepned panel
+          });
+        }
+        else{
+          //if no hash is called, open first panel
+          $('.panel-group .panel:first-child .panel-collapse').addClass('in');
+          $('.panel-group .panel:first-child .panel-title a.no-scroll i.fa').addClass('panel-isOpen');
+        }
+      };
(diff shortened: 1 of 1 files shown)


assets/js/main.js
```diff
@@ -387,27 +387,28 @@ $(function(){
...
-      });
-    }
-    else{
-      //if no hash is called, open first panel
-      $('.panel-group .panel:first-child .panel-collapse').addClass('in');
-      $('.panel-group .panel:first-child .panel-title a.no-scroll i.fa').addClass('panel-isOpen');
-    }
-  };
-  activateAccordion(getHash);
+  if(location.pathname.indexOf('/guides') === -1) {
```
Etiene: What happens if we add new pages also without this accordion?

hngerebara: I think we can then take off the condition as it will no longer throw an error. Another thought I had was rather what if we have other links that have guides somewhere and has `panel` class. This would be a problem. So I have just updated the code to rather check the value after the base url

hngerebara:
The issue is that such a whitelist at this point will be almost all the page that existed pre-deployment guides work. Seems like having a panel class has been the default up until now

brikis98:
It's not a whitelist, really. You search for the class name and see if anything has `.panel` on it and only try to apply the behavior if something matches.

hngerebara:
https://drive.google.com/file/d/1a904Xeu_7-seirBs-os0x4aonWo8iXFP/view?usp=sharing

assets/js/main.js
```diff
@@ -387,27 +387,31 @@ $(function(){
...
-      $panelHeight = accordion.parents('.panel').height();
-      console.log('Offset Top: ' + accordion.parents('.panel').offset().top);
-      $('html,body').animate({
-          scrollTop: accordion.parents('.panel').offset().top //scroll to accordion
-      }, 100, function () {
-          accordion.click(); //simulate click
-          accordion.parents('.panel').find('.panel-title a.no-scroll i.fa').addClass('panel-isOpen'); //point caret down for oepned panel
-      });
+    const accordion = $('a[href="' + id + '"]');
+    const $panelHeight = accordion.parents('.panel').height();
```
brikis98: Won't `.height()` blow up if there's nothing that matches `.panel`? It seems like the check you want is `accordion.length > 0`.

hngerebara: I don't think that would be the right conditional in this case. `accordion.length > 0` will be true in this case and will still throw an error when it gets to the panel height. I noticed the code that retrieves the parents($('.panel'))  always returns something. Not exactly sure where the panel class is in this case but the `panel` class isn't the actual issue. I tried this $('a[href="' + location.hash + '"]').parents($('.panel'))  in my console for https://gruntwork.io/guides/#kubernetes/ and it returns an output . The error is coming from .height()of the panel returning null for some pages that's why I'm using that condition

brikis98: Check both `accordion.length > 0` and `accordion.parents('.panel').length > 0`.

hngerebara: That's true. Not sure why I never thought of that

assets/js/main.js
```diff
@@ -387,26 +387,29 @@ $(function(){
...
-    else{
-      //if no hash is called, open first panel
-      $('.panel-group .panel:first-child .panel-collapse').addClass('in');
-      $('.panel-group .panel:first-child .panel-title a.no-scroll i.fa').addClass('panel-isOpen');
-    }
-  };
+    const accordion = $('a[href="' + id + '"]');
+
+    //Check panel class exists 
+    if(accordion.length > 0 && accordion.parents('.panel').length > 0){
```
brikis98: Can you fix the formatting on this file? Space after `if`, space before `{`.

assets/js/main.js
```diff
@@ -387,26 +387,29 @@ $(function(){
...
+      if (id.length)//check if hash isn't empty
+      {
```
brikis98: Formatting here too... Curly brace should be on prev line

assets/js/main.js
```diff
@@ -387,26 +387,29 @@ $(function(){
...
+        $('html,body').animate({
+            scrollTop: accordion.parents('.panel').offset().top //scroll to accordion
+        }, 100, function () {
+            accordion.click(); //simulate click
+            accordion.parents('.panel').find('.panel-title a.no-scroll i.fa').addClass('panel-isOpen'); //point caret down for oepned panel
+        });
+      }
+      else{
```
brikis98: More formatting, should be `} else {`...

hngerebara: Ahhhhh I will get IntelliJ now. My code editor is really messing me up

PR #187, opened 2019-09-19
Switch footer text to "Deployment Guides"

brikis98:
Otherwise, it ends up wrapping in the footer, which is confusing:

Screen Shot 2019-09-19 at 10 01 48 PM


brikis98:
Trivial change, so merging immediately.

PR #188, opened 2019-09-19
Rename "Service Catalog" to "Infrastructure as Code Library"

brikis98:
We haven't launched the new Service Catalog yet, with the new design, so it doesn't make sense to start referring to it as the Service Catalog in some places and not others. We'll switch to the new terminology everywhere later on.


brikis98:
Merging so I can get this stuff live. Feedback welcome.

PR #189, opened 2019-09-19
Add cross-links to deployment guides

brikis98:
This allows them to be found from the rest of our website


brikis98:
This is just adding a few links, so hopefully no one will object too much if I merge and move along!

PR #191, opened 2019-09-24
add padding to post detail bottom

hngerebara:
**Add paading to post detail bottom**
Screenshot 2019-09-24 at 12 02 29 PM
Screenshot 2019-09-24 at 12 02 37 PM

### DIFF
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -2065,6 +2065,7 @@ body .guides .guides-section-white {
 
 // Post Detail styles
 .post-content {
+  padding: 0;
   color: @guides-font;
   line-height: 1.38;
   padding-bottom: 50px;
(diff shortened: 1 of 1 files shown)


assets/css/global.less
```diff
@@ -2065,9 +2065,9 @@ body .guides .guides-section-white {
 
 // Post Detail styles
 .post-content {
-  padding: 0;
   color: @guides-font;
   line-height: 1.38;
+  padding-bottom: 50px;
```
brikis98: Removing the `padding: 0` above will cause other changes on the page we might not want. Perhaps override solely the bottom padding?

hngerebara: Okay

PR #194, opened 2019-09-24
SEO RFC

Etiene:


rfc/seo.md
```diff
@@ -0,0 +1,72 @@
+# Gruntwork.io SEO RFC
+
+I read a lot of material in SEO, including going through a [guide to SEO](https://moz.com/beginners-guide-to-seo/).
+
+## Important take aways
+
+- We currently have some broken stuff in our website that should take priority over trying to improve SEO by creating more content. Ranked by (my perceived) importance:
+  - Searching for site:gruntwork.io/guides will show us that our content is duplicated between the www and non www versions of the website. This is diluting our ranking. We need to add a canonical link to our headers
```
brikis98: Doh.

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+  - Our posts don't have meta descriptions
+  - Our internal links usually have the same anchor text (the title of the guide), but Google interprets that as not being natural and will penalize us for that
```
brikis98: Really? It used to be that you put into the link the _exact_ text you want to be ranked for... I wonder if this is some new rule?

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+  - We don't have a [sitemap](https://support.google.com/webmasters/answer/183668?hl=en)
```
brikis98: https://gruntwork.io/sitemap.xml

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+- Many websites use rel=nofollow (blocks crawling and backlink ranking bonus) on all external content or on all of their comment section. This is so because linking to bad quality content penalizes their own ranking. We need to verify if it's the case before engaging in any comment marketing for SEO reasons since this nullifies SEO advantages (but might still be useful for driving users that will click on it). Backlinks are also more important from topically similar websites. Link in hashicorp >>>> link in reddit.
```
brikis98: Yea, rel=nofollow is definitely common... Not sure if there's an easy way to add that for links created via Markdown... Perhaps punt on that until we migrate the site to SaaS?

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+- All links on a page have a [link equity](https://moz.com/learn/seo/what-is-link-equity) and our own internal links are competing for this equity. We need to add nofollow to any link that we think should not earn a share. Attention: we can be penalized for linking to bad quality content but also for linking too much to some content! We should dive deeper into this and see if this means we should add nofollow to most of our AWS docs links.
+- We could be more intentional about our robots.txt. Not crawling irrelevant pages or low performing pages raises the average performance of the website, which raises the ranking of the website.
+- Landing pages might not be a good idea after all. Thin content used to be good SEO practice (have multiple pages for each thing), but this does not have lots of advantages to users so now Google penalizes this. As a rule of thumb, what's good for users will be good for rankings since Google uses machine learning on user behavior. So we should watch things like time on page, bounce rate etc. identify low performing pages vs high performing pages and perhaps remove or improve low performing pages.
```
brikis98: I think this would be good for users if users were searching for some terms exactly. E.g., A landing page for "Docker orchestration tools" (in the service catalog) seems like a likely search term and a useful page to have. Harder to say if, say, a landing page for "Kubernetes deployment guides" is as valuable—i.e., if users ever search "deployment guide." Maybe "tutorial" or "walkthrough" is more common?

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+
+- According to the poing above, user engagement becomes one of the most important things on a website. Which means adding the comment section is a good idea.
+
+- Keywords are important but sometimes the most authoritative keywords are not the best ones. High-volume keywords are also more competitive to rank higher (keyword difficulty). Carefully [exploring the long tail](https://moz.com/blog/long-tail-seo-target-low-volume-keywords-whiteboard-friday) of relevant keywords to us might offer us a better advantage in ranking and in financial conversions.
+
+  > For example, a person searching for "shoes" is probably just browsing. On the other hand, someone searching for "best price red womens size 7 running shoe" practically has their wallet out!
+
+- Perhaps we were not intentional enough concerning keywords in our current content. According to the point above, phrases formulated as questions are SEO gold. We could do a search on the most common questions on our space and make sure these questions are phrased in our content
```
brikis98: +1

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+
+- Writing content is HARD so we need to make sure we are making the most for our buck. We need a strategy before deciding which content to write about. I've seen these two content strategies which seem very interesting
```
brikis98: I think our #1 driver for what content we write should be guides that customers are asking for. E.g., we had _many_ customers get stuck on the basic account + IAM setup. And we'll have many more who need help with K8S services, secrets management, CI / CD, and a handful of other topics. So, even if it might not perform as well SEO-wise, I still want that to be our #1 driver.

However, once we've settled on the content we want to write, the other ideas in here seem like an excellent way to get the most bang for the buck from that content. Also, once we have the core guides in place and start to approach the long tail, searching for "10x content" is also a great idea.

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+  - 10x content (😂 geez I *hate* the name)
+    - Search the keyword(s) you want your page to rank for
+    - Identify which pages are ranking highly for those keywords
+    - Determine what qualities those pages possess
+    - Create content that’s better than that
+  - Top content refurbishment
+    - See which of your content performs better
+    - Take that content and [refurbish it for other platforms](https://moz.com/blog/refurbishing-top-content-whiteboard-friday) (Slideshare, YouTube, Instagram, Quora, etc.) to expand acquisition funnel beyond Google.
+
+- Page speed is a very important SEO metric so we need to be extremely careful when making the website dynamic. Perhaps it's best to never migrate the guides to be dynamic? But images are the usually the top 1 culprit in page speed. We should consider optmizing the images for mobile and compressing all the images we use.
```
brikis98: Good to track! Since deployment guide content never changes, we can either always serve it statically _or_ put a CDN in front of the website with a 24h cache, which is likely the fastest of all.

rfc/seo.md
```diff
@@ -0,0 +1,72 @@
...
+
+- There seems to be no correlation with text length and rankings (I looked this up to see if it is a problem that our guides are very long and if maybe we should break them up, but apparently it doesn't matter)
+
+- Breadcrumbs are SEO-useful if they contain [micro data](http://schema.org/breadcrumb)
+
+- Google ranks accessibility and mobile-friendliness
+
+- Trustworthiness goes a long way both for search **and for users**. I think we need to make more prominent on our website that we maintain terratest, terragrunt & that jim wrote terraform up and running.
```
brikis98: How is trustworthiness measured by Google?

rfc/seo.md
Etiene: Yes and no! 
We do want the text we want to rank to be represented. But the question here is
```
<a href="...">How to deploy a production-grade Kubernetes cluster on AWS</a>
<a href="...">How to deploy a production-grade Kubernetes cluster on AWS</a>
<a href="...">How to deploy a production-grade Kubernetes cluster on AWS</a>
```
vs.
```
our <a href="...">guide on deploying dockerized containers on AWS</a>
you can <a href="...">learn how to configure a Kubernetes cluster</a>
we have a <a href="...">whole deployment walkthrough for EKS</a>
```

Quoted from the SEO guide I followed:
>Too many internal links using the same, keyword-stuffed anchor text can appear to search engines that you’re trying to manipulate a page’s ranking. It’s best to make anchor text natural rather than formulaic.

rfc/seo.md
Etiene: oops!

rfc/seo.md
Etiene: > I think our #1 driver for what content we write should be guides that customers are asking 

I agree 100%! My intention on bringing this up is for the case where we have more than one request or possibility and this could help us pick which one we write first and ways to tweak it to rank higher

rfc/seo.md
Etiene: CDN would be a great call, yes

rfc/seo.md
Etiene: backlinking by authorithative websites (we already know about this, so I didn't go into details here), but it got me thinking that users probably have the same concerns

Etiene: A good chunk of our users come to us exactly because they found about us from these sources. But now that we are targeting more organic approaches, they might not be aware.

rfc/seo.md
Etiene: Having a look at it now, it looks like it could be improved.

For example, this doesn't seem like it should be crawled 
```
<url><loc>https://www.gruntwork.io/assets/css/bootstrap/js/tests/visual/carousel.html</loc><lastmod>2018-10-30T22:34:51-07:00</lastmod></url>
```

Etiene:
> Agreed on your overall conclusion of not obsessing over tags. However, I also want to time cap our SEO efforts and not go too crazy here. Do you think you could take a guess at a few very high leverage SEO improvements we could do in, say, a budget of 3 days?

1. The canonical urls absolutely need to be fixed. 
2. Maybe taking the website on a tour of those tools to see if anything else really important pops up.
3. Then I'd go for the rest of suggested approach on a more flexible order and stop when it reaches 3 days?

rfc/seo.md
Etiene: Markdown accepts plain html just fine. We could just use the old fashioned `<a>` but yes, I'm not sure how much of a priority this is. It requires a bit more of research to understand the depth of the problem. But it's good to be aware. I didn't about link equity before so I really wanted to call it out.

Etiene:
Something else I forgot to mention on the RFC is `https://gruntwork.io/guides/kubernetes`.
Categories definitely take priority over tags because they are represented in the URL structure of the website. This is a problem to users and crawlers, but I dont't quite get the depth of the problem yet.

So I'd rework the comment above to be:

1. The canonical urls absolutely need to be fixed
2. Do a dive on 404 intermediary urls 
3. Taking the website on a tour of those tools to see if anything else really important pops up
4. Reprioritize everything that's left then go for the rest of suggested approach on a more flexible order and stop when it reaches 3 days?

rfc/seo.md
brikis98: Good to know, thx!

brikis98: (BTW, consider adding this example to the RFC)

rfc/seo.md
brikis98: Agreed!

rfc/seo.md
brikis98: Roger

brikis98:
Agreed! Do you want to take the lead on this stuff? Seems like a good set of tasks to tackle in your final full-time week.

mcalhoun:
closing stale PR

PR #198, opened 2019-09-30
Implement comments

hngerebara:
Implement a commenting system for the guides using [utteranc](https://utteranc.es/)
Can also share on the various social media pages

- [x] Facebook
- [x] Reddit
- [x] Linkedin
- [x] Twitter
- [x] Hackernews 

Screenshot 2019-09-30 at 12 22 33 PM

Share on reddit
Screenshot 2019-09-30 at 12 24 19 PM

Share on twitter
Screenshot 2019-09-30 at 12 24 09 PM

Share on linkedin
Screenshot 2019-09-30 at 12 23 53 PM

Share on facebook
Screenshot 2019-09-30 at 12 23 31 PM

N/B
- I created a new repository [guides-comments](https://github.com/gruntwork-io/guides-comments) where all the comments will be saved 

- Still having issues implementing loading more comments. Etiene advised I create a new ticket to further investigate the issue rather than delay this PR. 

- I also implemented the sharing as a pop up. Is this okay or better to be open on a new tab?


Etiene:
I paired with hope on the *load more* thing, just to give a bit of context on the issues we went through:

1. Utterances currently has some problems on their pagination. It has a non-configurable page size of 25 comments which doesn't seem to be working correctly. It loads all comments regardless (we tried making an example post with more than 25 comments)
2. We could hide some of the comments after load and add a "load more" button, however that means we don't get any of the performance benefits since all the comments are loaded anyway. We could still do it just for the aesthetics however when pairing with hope on this, an important issue came up: the comments are hosted on github and are utterances shows this through an iframe, which means that attempts to manipulate it get flagged for XSS `SecurityError: Permission denied to access property "document" on cross-origin object`. 

So I suggested hope to work on the rest of the work and make a PR already and discuss the possibility of separating the "load more" as a new card

brikis98:
Hm, that sounds like quite a few limitations. Before we go too far down the path of yak shaving, have you tried posting a bug with the utterances folks? Are they responsive? If we're having issues so early on, are we better off trying a different commenting system after all?

hngerebara:
Was supposed to create the issue yesterday but had network constraints. I just created an [issue](https://github.com/utterance/utterances/issues/205)

_includes/comments.html
```diff
@@ -0,0 +1,60 @@
...
+        <div class="col-md-3">
+            <div class="center">
+                <h2 class="no-anchor">Comments </h2>
+            </div>
+            <div class="center share-buttons">
+                <ul>
+                    {% capture url %}{{site.url}}{{page.url}}{% endcapture %}
+                    <li>
+                        <a  href="https://www.facebook.com/sharer/sharer.php?u={{ url }}"
+                        onclick="window.open(this.href, 'pop-up', 'left=20,top=20,width=500,height=500,toolbar=1,resizable=0'); return false;"
```
brikis98: Will this be blocked by pop-up blockers? How does it compare to `target="_blank"`?

_includes/comments.html
```diff
@@ -0,0 +1,60 @@
...
+    <div class="center">
+        <div class="col-md-3">
+            <div class="center">
+                <h2 class="no-anchor">Comments </h2>
+            </div>
+            <div class="center share-buttons">
+                <ul>
+                    {% capture url %}{{site.url}}{{page.url}}{% endcapture %}
+                    <li>
+                        <a  href="https://www.facebook.com/sharer/sharer.php?u={{ url }}"
```
brikis98: What about page title and description? Those can be pre-filled in the tweet/FB share/reddit share/etc too, right?

brikis98: Also, I think you will need to URL-encode those parameters. E.g., `u={{ url | url_encode }}`.

_includes/comments.html
```diff
@@ -0,0 +1,60 @@
...
+            </div>
+        </div>
+    </div>
+    <script src="https://utteranc.es/client.js"
+        repo="gruntwork-io/guides-comments"
+        issue-term="pathname"
+        theme="github-light"
+        crossorigin="anonymous"
+        async>
+    </script>
```
brikis98: What a lovely, clean API 👍 

So was the only hold-up with utterances that pagination didn't seem to work?

_includes/comments.html
hngerebara: No it doesn't get blocked by pop-up blockers. But I'll also prefer opening in a different tab so the user gets to properly see how it looks before sharing. I wasn't sure how best to implement it.

_includes/comments.html
hngerebara: Yeah but now works. We needed to have at least 53 comments before pagination can work

hngerebara:
Update
-------
Now opening sm share in a new tab
Pagination now works

_includes/comments.html
brikis98: My guess is that a an `<a target="_blank">` is a better fit here than `window.open`. Make sure to add `rel="noopener noreferrer"` too.

_includes/comments.html
```diff
@@ -0,0 +1,50 @@
...
+                            <i class="sm-icon fa fa-reddit-alien"> </i>
+                        </a>
+                    </li>
+                    <li>
+                        <a  href="https://www.linkedin.com/shareArticle?mini=true&url={{ url | url_encode }}" target="_blank" title="Share on LinkedIn" >
+                            <i class="sm-icon fa fa-linkedin"></i>
+                        </a>
+                    </li>
+                    <li>
+                        <a  href="https://twitter.com/intent/tweet?text={{ page.title }}&url={{ url | url_encode }}" target="_blank" title="Share on Twitter" >
```
brikis98: Use `url_encode` on `page.title` and ALL other URL params

_includes/comments.html
brikis98: For context, see https://developers.google.com/web/tools/lighthouse/audits/noopener

PR #199, opened 2019-09-30
zoom images on hover

hngerebara:
Implement zooming images on hover
https://drive.google.com/file/d/11N9pwkBK-VoP30Hzzu8I-FntcMfJu_3C/view?usp=sharing

My concern with this approach:
A user might just be reading the post but on hover of an image it zooms out which isn't the actual intention of the user. A better approach is usually including a checkbox on the image such that when the user clicks on the image, it is then zoomed. Since the posts are automatically loaded, I will have to use js to include a checkbox to the image. 
Will love to get better ideas on how to implement this.

Thanks

### DIFF
--- a/assets/css/global.less
+++ b/assets/css/global.less
@@ -2118,11 +2118,7 @@ body .guides .guides-section-white {
     display: block;
     margin: 0 auto;
     &:hover {
-      -webkit-transform:scale(1.25); /* Safari and Chrome */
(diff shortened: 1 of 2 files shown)


brikis98:
Hmm, you're right, zooming on hover is definitely annoying. Perhaps we add JS code to make each image clickable and open the full size image in a new tab?

PR #201, opened 2019-09-30
Only load prism css on guides instead of all pages at gruntwork.io

Etiene:
1. does not use same layout of default.html for post.html
2. copy pastes contents of default.html around post.html
3. removes prism.css from head 
4. adds prism.css to post.html

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -1,49 +1,59 @@
 ---
-layout: default
 custom_js:
 - prism
 - guides
 ---
-
-<div  class="post-bg">
-  <div class="container-fluid">
-      <div class="post-detail row equal">
-        <div class="col-md-2-5">
-          <div id="toc" class="js-scroll-with-user" data-scroll-after-selector=".navbar-default" data-scroll-until-selector=".guides-section-white">
-              {% toc %} {{page.document | tocify_asciidoc}}
-          </div>
-        </div>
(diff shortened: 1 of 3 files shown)


_layouts/post.html
```diff
@@ -1,49 +1,59 @@
...
-          </div>
+<!doctype html>
+<html lang="en">
+    <head>
+        {% include head.html %}
+        <link rel="stylesheet" href="{{ '/assets/css/prism.css' | prepend: site.baseurl }}"/>
+    </head>
+    <body{% if page.slug %} class="{{ page.slug }}{% if page.slug != 'index-page' and page.slug !='how-it-works' and page.slug != 'guides' %} sub-page{% endif %}" {% endif %}>
+        {% include navbar.html %}
+        <div  class="post-bg">
```
Etiene: tab

Etiene:
> What's the advantage of duplicating the layout code?

So `post` and `default` can have different `<head>` contents, one with and one without prism. Before there was one single layout, `default`, with prism.

brikis98:
> So post and default can have different <head> contents, one with and one without prism. Before there was one single layout, default, with prism.

Can we make CSS conditional, like JS was? So, one layout, but an if-statement/loop to add custom CSS on some pages?

brikis98:
Just noticed this PR was never merged?

mcalhoun:
closing stale PR

PR #208, opened 2019-10-07
fix social media share thumbnail

hngerebara:
I was not able to test sharing facebook sharing locally. I also tried using ngrok, but to no avail. I'm guessing this should work as the `og:image` and `twitter:image` url no longer points to the fixed thumbnail that always displays when we share

**Before:**
Screenshot 2019-10-07 at 1 02 23 PM

**After**
Screenshot 2019-10-07 at 12 56 18 PM

### DIFF
--- a/_includes/share-meta.html
+++ b/_includes/share-meta.html
@@ -1,17 +1,19 @@
 {% capture title %}{% if page.title %}{{ page.title }}{% else %}{{ site.name }}{% endif %}{% endcapture %}
 {% capture description %}{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines }}{% else %}{{ site.description }}{% endif %}{% endcapture %}
+{% capture img %}{% if page.image %}{{ page.image }}{% else %}{{ site.thumbnail_path }}{% endif %}{% endcapture %}
 
 <meta name="twitter:card" content="summary" />
 <meta name="twitter:site" content="{{ site.data.contact.twitter.username }}" />
 <meta name="twitter:title" content="{{ title }}" />
 <meta name="twitter:url" content="{% include canonical-url.html %}" />
 <meta name="twitter:description" content="{{ description }}" />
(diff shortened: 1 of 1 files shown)


_includes/share-meta.html
```diff
@@ -1,17 +1,19 @@
...
-<meta name="twitter:image" content="{{ site.url }}{{ site.thumbnail_path }}" />
+<meta name="twitter:image" content="{{ site.url }}{{ img }}" />
 
 <meta property="og:title" content="{{ title }}"/>
 <meta property="og:url" content="{% include canonical-url.html %}"/>
 <meta property="og:site_name" content="{{ site.name }}"/>
 <meta property="article:author" content="{{ site.data.contact.facebook.url }}"/>
-<meta property="og:description" content="{{ description }}"/>
-<meta property="og:image" content="{{ site.url }}{{ site.thumbnail_path }}"/>
+<meta property="og:description" content="{{img}}"/>
```
brikis98: Why would `description` become `img`?

_includes/share-meta.html
```diff
@@ -1,17 +1,19 @@
 {% capture title %}{% if page.title %}{{ page.title }}{% else %}{{ site.name }}{% endif %}{% endcapture %}
 {% capture description %}{% if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines }}{% else %}{{ site.description }}{% endif %}{% endcapture %}
+{% capture img %}{% if page.image %}{{ page.image }}{% else %}{{ site.thumbnail_path }}{% endif %}{% endcapture %}
```
brikis98: Do we set `page.image` on the deployment guides?

_includes/share-meta.html
hngerebara: Ooh no I changed that but didn't push the update

_includes/share-meta.html
hngerebara: Yes each deployment guide has an image that can be accessed with `page.image`

PR #213, opened 2019-10-13
Fix VPC guide links

brikis98:


brikis98:
This is just a URL tweak, so I feel confident in merging and deploying.

PR #217, opened 2019-10-15
Clean js file

hngerebara:
- Clean up Search file.
- Replace let with cont where needed.
- Split code into functions

N/B Searching and filtering still works on both IAC page and guides page

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -23,13 +23,13 @@
   function debounce(func, wait, immediate) {
     let timeout;
     return function () {
-      let context = this,
+      const context = this,
         args = arguments;
-      let later = function () {
+      const later = function () {
         timeout = null;
         if (!immediate) func.apply(context, args);
       };
-      let callNow = immediate && !timeout;
+      const callNow = immediate && !timeout;
       clearTimeout(timeout);
       timeout = setTimeout(later, wait);
       if (callNow) func.apply(context, args);
@@ -48,7 +48,7 @@
         $('#search-results-count').text("0 repos");
       }
     } else {
-      if (totalCount > 0 ) {
+      if (totalCount > 0) {
         $('#search-results-count').html("<strong>" + totalCount + "</strong> post(s) found");
       }
     }
@@ -68,7 +68,7 @@
    */
   function showInitialItemsCount(searchEntry) {
     let numSubmodules = 0;
-    for (let i = 0; i < searchEntry.entries.length; i++) {
+    for (const i = 0; i < searchEntry.entries.length; i++) {
       numSubmodules += libraryEntries[i].num_submodules;
     }
     showItemsCount(searchEntry, searchEntry.entries.length, numSubmodules);
@@ -79,22 +79,8 @@
    */
   function detectSearchEntry() {
     let entries = [];
-    if (window.libraryEntries) {
-
-      entries = window.libraryEntries;
-      return {
-        entries,
-        type: 'libraryEntries'
-      };
-
-    } else if (window.guideEntries) {
-
-      entries = window.guideEntries;
-      return {
-        entries,
-        type: 'guideEntries'
-      };
-    }
+    entries = window.libraryEntries ? window.libraryEntries : window.guideEntries;
+    return entries;
   }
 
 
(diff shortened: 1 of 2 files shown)


assets/js/search.js
```diff
@@ -39,62 +39,66 @@
   /**
    * Function displays the total items displayed on the page
    */
-  function showItemsCount(searchEntry, totalCount, numSubmodules) {
-    if (searchEntry.type === 'libraryEntries') {
+  function showItemsCount(totalCount, numSubmodules) {
+    if ($('.table-clickable-row'.length > 0)) {
```
brikis98: This looks like a bug... Shouldn't it be `$('.table-clickable-row').length > 0`? That is, check the paren placement.

assets/js/search.js
```diff
@@ -39,62 +39,66 @@
...
    * Function to display all items on the page
    */
   function showAllItems() {
-    $('.guide-card').show();
-    $('.category-head').show();
-    $('.categories ul li').show();
+    return (
+      $('.guide-card').show() &&
+      $('.category-head').show() &&
+      $('.categories ul li').show()
```
brikis98: Why change to `&&` here and below?

assets/js/search.js
```diff
@@ -103,99 +107,87 @@
...
         $('#search-results-count').hide();
         $('#no-matches').show();
         return;
       }
 
-      showItemsCount(searchEntry, matches, submoduleMatches);
-    } else {
-      if (searchEntry.type === 'libraryEntries') {
+      showItemsCount(matches, submoduleMatches);
+    } else if ($('.table-clickable-row'.length > 0)) {
```
brikis98: Same here

assets/js/search.js
```diff
@@ -103,99 +107,87 @@
...
 
         //Checks if results were found and displays results accordingly
         if (matchesAll) {
           displayCategory(entry);
-          $("#" + entry.id).show();
+          $(`#${entry.id}`).show();
           matches++;
-          (searchEntry.type === 'libraryEntries') ? submoduleMatches += entry.num_submodules: submoduleMatches = 0;
-          return;
+          $('.table-clickable-row'.length > 0) ? submoduleMatches += entry.num_submodules: submoduleMatches = 0;
```
brikis98: Same here

brikis98: Also, ternary syntax is primarily useful as an expression that returns a value, and you're trying to pick between two possible values in a one-liner. Here, you're using it for side effects, which seems harder to read without much gain. Consider turning into normal if/else.

assets/js/search.js
```diff
@@ -39,62 +39,66 @@
...
-    } else if (window.guideEntries) {
-
-      entries = window.guideEntries;
-      return {
-        entries,
-        type: 'guideEntries'
-      };
-    }
+    entries = window.libraryEntries ? window.libraryEntries : window.guideEntries;
+    return entries;
```
brikis98: You can collapse these three lines into one: `return window.libraryEntries ? ... : ...;`

assets/js/search.js
hngerebara: I felt using `&&` was better since I was returning. Is it better to terminate each statement instead?

assets/js/search.js
hngerebara: Definitely a bug. Quite surprised why it worked without throwing any errors. I'll fix it

hngerebara:
Yes, we should definitely pair. I will love to hear your thoughts. I've created a Jira ticket [here](https://gruntwork.atlassian.net/browse/HSTN-624)

hngerebara:
This is ready for review again @brikis98

assets/js/search.js
```diff
@@ -14,188 +14,169 @@
...
   }
 
   /**
    * Function to display all items on the page
    */
   function showAllItems() {
-    $('.guide-card').show();
-    $('.category-head').show();
-    $('.categories ul li').show();
+    return ($('.guide-card').show() && $('.category-head').show() && $('.categories ul li').show())
```
brikis98: How is this return value used?

PR #218, opened 2019-10-15
Generated updated TOS .docx

bwhaley:


josh-padnick:
Was this docx file actually updated? Jekyll generates a new one on every build right now but we generally only upload a new one when the underlying terms change.

bwhaley:
Yes, I generated it to capture the 15 => 30 day update.

josh-padnick:
Ah roger, thanks!

PR #219, opened 2019-10-17
Disable scroll to position

hngerebara:
Remove link from parent navigation item so users can view content within a block before deciding if they want to scroll to position or not

https://drive.google.com/file/d/14vA8EXzSwj-bQQMCR1CmHoN_1p7xaPh8/view?usp=sharing


hngerebara:
It is automatically generated by Jekyll toc. It can only be removed via Js

assets/js/guides.js
```diff
@@ -120,4 +120,19 @@ $(document).ready(function () {
     ga('send', 'event', location.pathname, 'subscribe-cta');
   })
 
+  //Remove link from parent navigation item and add classes when clicked
+  $.each($("#toc ul > li"), function (index, obj) {
```
brikis98: Minor nit: I think `$("#toc ul > li").each` is more idiomatic.

assets/js/guides.js
```diff
@@ -120,4 +120,19 @@ $(document).ready(function () {
...
+    if ($(obj).has("ul").length) {
```
brikis98: Minor nit: `.length > 0` is more readable and relies less on magical truthiness in JS. Also, store the output of `$(obj).has("ul")` in a variable so it can be reused below instead of copy/pasted.

assets/js/guides.js
```diff
@@ -120,4 +120,19 @@ $(document).ready(function () {
...
+      const parentNavItem = $(obj).has("ul");
+
+      parentNavItem.find("a:first").removeAttr("href");
```
brikis98: Why remove? Seems like you're already overriding the click behavior below anyway.

hngerebara: Because on click, if the link still exists, it scrolls to that position. So I'm removing the link while clicking removes existing classes

PR #220, opened 2019-10-18
Update terraform training author to include yori

yorinasub17:
With the tf12 refresh, the training course now includes me as an author so adding that in.


yorinasub17:
Thanks for review! Merging this in!

josh-padnick:
@yorinasub17 Did you want to deploy this as well?

yorinasub17:
Sorry for the late reply. Yes I did want to deploy this, but got busy. Looks like it was deployed as a part of https://github.com/gruntwork-io/gruntwork-io.github.io/pull/221

PR #221, opened 2019-10-22
Update the Deploying a Dockerized app on GKE guide to TF 0.12

robmorgan:
This PR updates the 'Deploying a Dockerized app on GCP and GKE' guide to support Terraform 0.12.


robmorgan:
@yorinasub17 thanks 👍

PR #223, opened 2019-10-28
Fix two typos in IaC Lib guide

brikis98:


brikis98:
Trivial change, so merging and deploying right away.

PR #224, opened 2019-10-29
Update job description

josh-padnick:
Sanity checked in https://gruntwork-io.slack.com/archives/CETJJ83EZ/p1572387973000500.


josh-padnick:
Merging now, but feedback welcome as always.

PR #236, opened 2019-11-27
fix what youll learn design[HSTN-746]

hngerebara:
Fix Styling
Before:
Screenshot 2019-11-27 at 3 17 23 PM

After:
Screenshot 2019-11-27 at 3 11 31 PM

### DIFF
--- a/assets/css/guides.less
+++ b/assets/css/guides.less
@@ -1079,7 +1079,7 @@ body .guides .guides-section-white {
       dt {
         margin-top: 15px;
       }
-    }
+	}
 
     pre.highlight > code, pre:not(.highlight) {
       padding: 15px;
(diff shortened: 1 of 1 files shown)


assets/css/guides.less
```diff
@@ -1091,6 +1091,17 @@ body .guides .guides-section-white {
       background: transparent !important;
     }
   }
+
+  h3#what-youll-learn-in-this-guide ~ .dlist {
```
brikis98: If you're already target an ID, why bother with the `h3`?

And what does the `~` selector do?

Also, should _all_ links in a definition list be styled this way? Or _only_ the ones in "what you'll learn in this guide" section?

hngerebara: Taken off the h3. 
~ is a sibling selector which I used to select `.dlist` class
Only what is within the `what you'll learn in this guide` section needs to be styled this way; that's why I'm selecting the id of that section

hngerebara:
@brikis98 Please review once more

assets/css/guides.less
brikis98: Roger

PR #239, opened 2019-11-29
[HSTN-741]fix newsletter input text field

hngerebara:
The video link attached captures the fix and the error currently happening on the website. 

https://drive.google.com/file/d/1w7RwUjVn7PQfRLo83JiyQxCSY1RazsBx/view?usp=sharing

When you enter a wrong email eg `te@gt` it redirects to the former newsletter page


assets/js/guides.js
```diff
@@ -137,4 +137,31 @@ $(document).ready(function () {
     }
   });
 
+  //Validate email
+  let validateEmail = email => {
+    let re = /@[\w.-]+\.[a-zA-Z]{2,6}$/;
```
brikis98: Generally, using a regex to validate email addresses is not a good idea. See:

https://davidcel.is/posts/stop-validating-email-addresses-with-regex/
https://stackoverflow.com/a/201378/483528

The validation you get from `<input type="email">` should be good enough... So what is the issue you're trying to solve?

hngerebara: For some reason, if we enter a wrong email like this `ghtx@gm` without the last part of an email, the form allows us to submit but redirects to the previous newsletter page. The one that had the purple background and there the error is thrown. I am trying to capture the error once the `subscribe` button is clicked.

brikis98: > For some reason, if we enter a wrong email like this ghtx@gm without the last part of an email, the form allows us to submit but redirects to the previous newsletter page.

It allows submit because the `<input type="email">` validation allows that form of an email address, where as MailChimp's email validation does not. So as a result, MailChimp's validation redirects you to a MailChimp-hosted page (something like https://gruntwork.us13.list-manage.com/subscribe/post?u=8ccaad0747ac13944cf318e0f&id=786bd365d1) to correct the email address. The design of that page is, IIRC, defined in MailChimp itself. So perhaps the proper fix is not to try to emulate MailChimp's email validation (which could change at any time in the future), but to update the design of that page?

hngerebara: There isn't so much that can be done but I changed the background color and the grunty image of the Mailchimp page to match what we currently use. Screenshot below
Screenshot 2019-12-03 at 9 18 37 PM

hngerebara:
Sam said the red border should only be displayed when the subscribe button is clicked hence the js code

assets/js/guides.js
eak12913: This definitely looks better than before but consider using this Grunty: https://gruntwork.io/assets/img/guides/layout/grunty-boxes.png as the one you've used has that box pattern behind him and I wonder if it's supposed to be part of a bigger picture so that it doesn't look cut off?

hngerebara: Done
Screenshot 2019-12-03 at 11 03 04 PM

brikis98: OK, that's a good improvement... But one issue: that red error color on top of the blue is very painful on the eyes ([it's a known issue with the human eye](https://www.blog.theteamw.com/2009/11/09/100-things-you-should-know-about-people-9-chromostereopsis-is-hard-on-the-eyes/)). 

Perhaps instead of the light blue background, we could go with the darker background on the gruntwork home page (`#1d2541`)? Or something else that contrasts better with red.

hngerebara:
New look
Screenshot 2020-01-07 at 9 58 13 AM

assets/js/guides.js
```diff
@@ -137,4 +137,25 @@ $(document).ready(function () {
...
+  $("#newsletter-subscribe").click(e => {
+    e.preventDefault();
+
+    let email = $("#guides-subscribe-email");
+    if (email.is(":invalid")) {
+      $("#guides-subscribe-email").css("border", "1px solid red");
+    } else {
+      $("#guides-newsletter-form").submit();
+    }
+    return false;
```
eak12913: Does this function need to return something?

hngerebara: Actually no it doesn't need to return anything. I've taken off the return statement

hngerebara:
@eak12913 This PR is still pending

_includes/newsletter.html
```diff
@@ -4,10 +4,10 @@
...
             <label>Newsletter</label>
             <p>Subscribe to the Gruntwork Newsletter to get notified when updates and new guides are released. No spam. Ever.</p>
-            <form  id="guides-newsletter-form"class="col-sm-10 col-sm-offset-1 col-xs-offset-1 col-xs-10" method="post" action="//gruntwork.us13.list-manage.com/subscribe/post?u=8ccaad0747ac13944cf318e0f&amp;id=786bd365d1" target="_blank">
+            <form  id="guides-newsletter-form" class="col-sm-10 col-sm-offset-1 col-xs-offset-1 col-xs-10" method="post" action="//gruntwork.us13.list-manage.com/subscribe/post?u=8ccaad0747ac13944cf318e0f&amp;id=786bd365d1" target="_blank">
                 <div class="center">
-                    <input type="email" name="EMAIL" class="email form-control guides-input-box" id="guides-subscribe-email" placeholder="Your Email Address">
-                    <input type="submit" class="btn guides-button" value="Subscribe">
+                    <input type="email" name="EMAIL" minlength="5" class="email form-control guides-input-box" id="guides-subscribe-email" placeholder="Your Email Address" required>
```
eak12913: Not a comment, just more of a question really: Why did you decide to add a minlength and why 5?

PR #240, opened 2019-11-29
fix button  stying

hngerebara:
The buttons on the website currently are not in uniform as specified in this [ticket](https://gruntwork.atlassian.net/browse/HSTN-743). This is a fix for it. Below are screenshots for the fixed buttons

 Screenshot 2019-11-29 at 2 54 02 PM

Screenshot 2019-11-29 at 2 53 48 PM


hngerebara:
That button should be white as well. She was talking about the size, font, and gradient which wasn't consistent

PR #241, opened 2019-12-02
[HSTN-744]Align outline

hngerebara:
Before:
Screenshot 2019-12-02 at 11 28 01 AM

After:
Screenshot 2019-12-02 at 11 27 28 AM


assets/css/guides.less
```diff
@@ -621,14 +621,14 @@ body .guides .guides-section-white {
...
+	top: 0 !important;
 }
 
 .bottom {
-	position: absolute;
-	bottom: 0;
-	top: auto;
+	position: absolute !important;
+    bottom: 0;
+    top: auto !important
```
brikis98: Why do these all need `!important`?

assets/js/guides.js
```diff
@@ -67,7 +74,7 @@ $(document).ready(function () {
     allNavLinks.removeClass('selected');
 
     // Only consider an item in view if it's visible in the top 20% of the screen
-    const buffer = $(window).height() / 5;
+    const buffer = $(window).height() / 7;
```
brikis98: Comment is now out of date

assets/js/guides.js
```diff
@@ -101,6 +108,8 @@ $(document).ready(function () {
           selectedNavLink.parent().addClass('expanded');
         }
       }
+    } else {
+      $('.sectlevel1 li a').first().addClass('selected');
```
brikis98: So if no headings are visible... We select the first one?

assets/js/guides.js
```diff
@@ -14,6 +14,13 @@ $(document).ready(function () {
     return element;
   };
 
+  //Move Outline to top of the position of first guide card
+  const startOutlinePosition = function () {
+    const firstGuideCardTop = $('.guide-card').first().offset().top;
+
+    $(".js-scroll-with-user").offset({top: firstGuideCardTop});
```
brikis98: Should the initial alignment not be done with CSS rather than JS? Otherwise, it'll jump on page load.

assets/js/guides.js
hngerebara: Yes, we want to have the first heading already active once we land on the page.

assets/js/guides.js
hngerebara: Nice point. I will do this in CSS.

assets/css/guides.less
hngerebara: I had to use important because the new position I specified on page load gets retained even when it should be picking up the fixed and bottom styles.

hngerebara:
Yeah. Sam filed it as a bug [here](https://gruntwork.atlassian.net/browse/HSTN-744). And I think it does look better than what we currently have

hngerebara:
Created a video [here](https://drive.google.com/file/d/1DTA0QRk_1TeH-KsLdCdmxu_8neP0UT2v/view?usp=sharing)

brikis98:
Thanks for the video. I see a few issues:

1. The nav now "jumps" to the top, rather than scrolling smoothly as the top of the browser gets to it.
1. If you scroll to the bottom and then back up, it also seems to "jump".
1. The highlighted nav item seems to be delayed? As in, you scroll to networking, but it only updates a little later... 

Overall, IMO, it's a more jarring experience than what we have now. Since this is a minor issue, I'd vote on punting on this until after Service Catalog is launched.

hngerebara:
This will be revisited once we finish with the information architecture. Seems FullstackHq already implemented it. Still tracking it in this jira [story](https://gruntwork.atlassian.net/browse/HSTN-875). @eak12913 should I close this or leave it open till we revisit it?

eak12913:
Let's close this PR and we'll re-open if we need it.

eak12913:
Closing this PR as we're going to be re-implementing these pages in the new info architecture.

PR #242, opened 2019-12-02
[HSTN-745]fix guide detail styling

hngerebara:
Fix design issue in [this](https://gruntwork.atlassian.net/browse/HSTN-745) jira card

Before:
Screenshot 2019-12-02 at 3 15 30 PM

After:
Screenshot 2019-12-02 at 3 14 51 PM


brikis98:
Hope, would you mind recording a video of what this behavior looks like? I want to better understand the following:

1. What does the page look like when it  first loads?
1. What happens as you scroll?
1. What happens when you click on an item?
1. What happens when you click the same item again?
1. What happens if someone gives you a direct link to some part of the page?
1. Does the updated design have enough padding? The bottom screenshot makes it look like the nav is really close to the left side of the page, which doesn't look great... But could also be because that's where the screenshot happens to cut off.

Thx!

hngerebara:
Made a video [here](https://drive.google.com/file/d/1MLwM2xPxJlYA9DcBYCi0wD5gFjJxJHy8/view?usp=sharing)

eak12913:
I see one issue: When you are scrolling down initially you see something like:

image

but just a moment later it changes to:

image

Seems like it should keep with the gray indicator and not change to the blue dot?

hngerebara:
No Sam what's it changed to the blue do. View [here](https://app.zeplin.io/project/5d581fcb1db0449b12b76763/screen/5d58283e91c8889afa13f5f2)

eak12913:
Oh - then perhaps it should always be that way? Otherwise I see a scenario where it's the gray version and THEN it changes to the dot. If that's as it was intended then feel free to disregard!

brikis98:
Thanks for the video Hope! I think it makes it clear that there are quite a few issues:

1. There isn't enough padding on top. The top of the browser ends up too close to the "Intro" text. See the current design (in prod) for the padding we should have.
1. The menu "jumps" when you first start scrolling down. That is, the top of the browser goes past it, and then a little while later, it catches up. The current design (in prod) scrolls smoothly as soon as the top of the browser approaches it.
1. I don't think it make sense to have something selected as soon as the page loads. It doesn't give you the feedback that the highlight scrolls with you until later, which looks odd. I've asked Sam this in https://gruntwork.atlassian.net/browse/HSTN-745.
1. As Eugene said, there's weird styling going on with the dot versus the caret:
    - On page load, the "Intro" text is bold and has a dot.
    - When you scroll past a random point, the dot becomes a caret (but nothing else has a dot).
    - Then the "Intro" text is no longer bold.
    - Then, suddenly, it goes back to both bold and with a dot. 
    - Similar issue happens with next nav item.
1. "What you'll learn in this guide" never got highlighted at all even though you scrolled past it.

I think our goal should be:

1. The nav text for the top-level sections  (e.g., "Intro", "Core Concepts") should be bold, blue, and with a dot when you are anywhere in that top-level section. When you are not in a top-level section, the nav text should be bold, gray, and with a caret.
1. The nav text for second-level sections (e.g., "What is Kubernetes?", "What you'll learn in this guide") should be bold, blue, with a dash when you are in that specific second-level section. When you are not in that second-level section, it should be normal (not bold), gray, and with a dash.
1. Keep the padding and scroll behavior as it is now (i.e., no "jumps").

Also, a sanity check: this doesn't seem like a critical bug to fix, especially compared to launching service catalog. Do you have service catalog tasks you could prioritize instead?

hngerebara:
Thanks, Jim. I'll have to punt this as I already have a service catalog task I'm working on.

hngerebara:
I have implemented it according to the 3 goals @brikis98 stated. Here is a [video](https://drive.google.com/file/d/1-ZOLH8YJkaaxVzJZnoHrFIaB2LQTGNom/view?usp=sharing)

brikis98:
Thanks Hope. That new video looks a lot better. The only odd things that remain are:

1. The nav on the left still seems to "hop" rather than slide down smoothly with scrolling. That is, at 0:01-0:02 in the video, the top of the page scrolls past the left nav, and then it "jumps" and is now scrolling with the page. Compare with what I see currently in prod:

    Kapture 2020-01-08 at 10 22 11
 
1. Similarly, the highlighting of nav items seems a bit delayed now too. E.g., at 1:36-1:38, you've scrolled past "Use EKS" and "VPC configuration", but it takes the left nav a little while to catch up.

Any idea what's causing the lag?

hngerebara:
Not really certain what's causing the lag locally. I tried using ngrok to expose the port and tried it with http://whatismyscreenresolution.net/multi-screen-test which doesn't seem to have any lag. 

In the meantime, it seems FullstackHq already implemented it. This will be revisited once we finish with the information architecture.  Still tracking it in this jira [story](https://gruntwork.atlassian.net/browse/HSTN-875). @eak12913 should I close this or leave it open till we revisit it?

eak12913:
Let's close this PR and we'll re-open if we need it.

eak12913:
Closing this PR as we're implementing these pages now in the new info architecture.

PR #244, opened 2019-12-05
Add testing talk to DevOps Resources

brikis98:


josh-padnick:
LGTM, but I wonder if people ever find this page. Is this worth a link from our DevOps Training Library?

brikis98:
> LGTM, but I wonder if people ever find this page. Is this worth a link from our DevOps Training Library?

I suspect very few. I think the proper solution is to add a full on free tier so that all visitors and customers see all our free and paid resources in training, code, and support.

Thx for the review. Merging and deploying now.

PR #249, opened 2019-12-23
Change iac page path to enable A/B testing against saas app page [HSTN-818]

oredavids:


josh-padnick:
Hey Ore, I'm not familiar enough with the A/B testing approach we're using to review this change, but we do want to start collecting our A/B testing data right away, so that while we're all relaxing on vacation, the data is accumulating. Therefore, if this is important to achieving that, and you feel reasonably confident this works as expected, I recommend you merge this and launch.

pages/infrastructure-as-code-library/index.html
```diff
@@ -21,9 +21,8 @@
       {% include_relative _sub-hero.html %}
       <div class="row row-library">
         <div class="col-xs-12">
-          {% include_relative _library.html %}
-          {% include_relative _features.html %}
-          {% include_relative _process.html %}
+          {% include_relative _library.html %} {% include_relative
+          _features.html %} {% include_relative _process.html %}
```
brikis98: Is this just a change in whitespace?

oredavids: Yes

PR #253, opened 2020-01-09
Fix link in footer

brikis98:
We noticed in Google Analytics that users were somehow going straight to the `/checkout` page without going through `/pricing`. Turns out we have the wrong link for the "Get Started" button at the bottom of every page.


brikis98:
Thx. Merging & deploying now.

PR #255, opened 2020-01-15
Remove Rob as a CODEOWNER

robmorgan:
I don't have much to do with this repo anymore so I'd like to remove myself as a CODEOWNER.


robmorgan:
@brikis98 yeah ok, I had a feeling we might.

PR #257, opened 2020-02-04
Point to module-security aws-config-multi-region instead of cis-compliance-aws aws-config

yorinasub17:
Since we migrated `aws-config` out of `cis-compliance-aws`, the guide should be updated to reflect that.


yorinasub17:
Thanks for review! Going to merge, but I don't have deployment permissions so can't push live.

brikis98:
Deployed!

PR #259, opened 2020-02-04
Update CODEOWNERS

eak12913:
Added myself to CODEOWNERS


brikis98:
@eak12913 The goal is to only have leads and contributors in the `CODEOWNERS` file. For the website, that should be: Matt, Eugene, Hope, Ore. This applies to all the repos where you're a lead or contributor. Could you update them accordingly?

PR #287, opened 2020-04-13
Fix Thank you page Redirect

eak12913:
This PR resolves the 404 that users of v1-static site were seeing after submitting the contact-us form. The link of the redirect should have been to `/static/contact`.


eak12913:
@josh-padnick - this should resolve the issues you reported in Slack.

PR #291, opened 2020-04-28
[HSTN-1321] Fix broken assets on static site

eak12913:
This PR addresses an issue where some assets like reference architecture
diagram as well as our terms of service and "why prod is down" downloads
were returning 404s.  The problem was the URL was pointing to
`/static/...` when in this case, it should not have been.


eak12913:
Given that this is currently breaking production, I'm merging the change. Comments welcomed. Thanks @oredavids for finding this issue!

PR #293, opened 2020-05-11
Initial draft of TFC/TFE integration content

bwhaley:
This update to the _How to use the Gruntwork Infrastructure as Code Library_ guide contains information about integration with TFE and TFC. Here are two direct links for easier reading:

* [Core concepts intro section](https://github.com/gruntwork-io/gruntwork-io.github.io/blob/tfc-tfe-integration/_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc#terraform-cloud-and-terraform-enterprise)
* [The nitty gritty details](https://github.com/gruntwork-io/gruntwork-io.github.io/blob/tfc-tfe-integration/_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc#integrating-with-terraform-cloud-and-enterprise)

The plan right now is to release this, then update and release the [corresponding blog post](https://medium.com/p/aca919ca92c2/edit). 

Open questions:

* Should we polish up the [prototype](https://github.com/gruntwork-io/prototypes/blob/terragrunt-tfc-demo/terragrunt-tfc-demo/terragrunt.hcl) and refer to that in this guide?
* If so, what changes should be made to the prototype?


_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+account that is only used for automation, and is not used by humans. In this case, the "machine" in question is the TFC
+executor. Create a new Github user, and email the username to mailto:support@gruntwork.io[support@gruntwork.io]. We'll make sure the user has access to our repositories.
+
+1. Next, add an SSH key to the new GitHub machine user. GitHub has https://help.github.com/en/enterprise/2.19/user/github/authenticating-to-github/connecting-to-github-with-ssh[easy-to-follow instructions].
+
+1. Now, add the private portion of the SSH key to TFC. You'll find the option under SSH Keys in the TFC organization settings. We called ours _Gruntwork access_.
+
+1. Set up a new workspace and link it to your code repository that uses Gruntwork's module as a source.
+
+1. In the Workspace settings, in the SSH key section, you can now choose the Gruntwork access key.
```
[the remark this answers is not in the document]
zackproser: I'm a huge fan of screenshots in these types of guides - so it's there's anywhere else it makes sense to add them for faster grokk-ability I say we go for it. It also helps people scanning the guide to figure out if this is what they're looking for or not.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+to deploy infrastructure with TFC and TFE. With a little more effort, you can also use Terragrunt! Throughout this
+section, to avoid repeating ourselves, we'll only mention TFC. The same patterns work just as well with TFE.
+
+[NOTE]
+.Obtain an API token
+====
+Before working with TFC/TFE from your local system, you'll need to obtain an API token. You can generate an API token in your user settings of the TFC UI. Follow https://www.terraform.io/docs/commands/cli-config.html#credentials-1[the official instructions] to use the token.
+====
+
+==== Using a Gruntwork module in a TFC workspace
```
brikis98, line 1429: I read through this section and to be honest, had a bit of trouble following it, even though me and you worked through this live just the other day!

I think (a) we need a bit more detail and (b) we should probably have more screenshots to guide the user here. There risk, of course, is that the TFC UI totally changes, and the screenshots are out of date, but I'd prefer a good experience now that needs updating later, rather than a crappy experience now and later.

Here's a rough outline of how I'd approach it:

**One-time setup**

1. Go to `app.terraform.io` (or your custom TFE URL).
1. Configure a VCS connection. Show screenshot creating an app in GitHub, then copying app details into TFC.
1. Add your SSH key. Show screenshot of associating SSH key with GitHub user and another copying the key into TFC.

**Deploy an example for learning**

(Cross-reference the examples, modules, test breakdown of our repos in this same doc)

1. Go to `app.terraform.io` (or your custom TFE URL).
1. Create workspace. (screenshot)
1. Pick VCS you added in initial setup and select a repo. We'll use package-sns (screenshot).
1. Enter a sub-folder. We want `examples/xxx`. (screenshot)
1. Enter variables. E.g. The name for the SNS topic (screenshot).
1. Enter env vars. E.g., AWS Access Keys and AWS region (screenshot).
1. Run `plan` (screenshot).
1. Run `apply` (screenshot). 
1. You have an SNS topic deployed! (screenshot)

**Deploy code for production**

1. Go to `app.terraform.io` (or your custom TFE URL).
1. Create a module in your `infra-modules` repo. E.g., Here's an example of a production SNS module.
1. Commit the changes: `git commit`, `git push`, etc.
1. Create workspace. (screenshot)
1. Pick VCS you added in initial setup and select a repo. This time, pick your `infra-modules` repo (screenshot).
1. Enter a sub-folder. We want `<path to your module in infra-modules>`. (screenshot)
1. Enter variables. E.g. The name for the SNS topic (screenshot).
1. Enter env vars. E.g., AWS Access Keys and AWS region (screenshot).
1. Run `plan` (screenshot).
1. Run `apply` (screenshot). 
1. You have an SNS topic deployed! (screenshot)

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+
+That's it! You can use this pattern for any of Gruntwork's Terraform modules. This even works for modules like github.com/gruntwork-io/terraform-aws-eks[`terraform-aws-eks`] which need to download external dependencies (like https://github.com/gruntwork-io/kubergrunt[`kubergrunt`]). Those modules use the special https://github.com/gruntwork-io/package-terraform-utilities/tree/master/modules/executable-dependency[`executable-dependency`] module to install the external dependency within the executor at run time.
+
+In the future, Gruntwork may offer modules through  a private Terraform registry to further simplify the process.
+
+[[terragrunt_tfc_tfe]]
+==== Using Terragrunt with TFC and TFE
+Configuring Terragrunt to work with TFC/TFE is a bit more involved. As mentioned previously, TFC and TFE execute
+`terraform` commands under the hood, and the executors that run these commands aren't able to execute the Terragrunt
+binary. Running Terragrunt directly through the TFC UI isn't an option.
```
brikis98, line 1486: I'd rephrase this slightly:

1. The TFC/TFE UI can only run the Terraform binary, whereas Terragrunt expects you to run it, and it, in turn, will run Terraform. 
1. Therefore, there's no way currently to use the TFC/TFE UI to trigger deployments.
1. However, you can use TFC/TFE to [run remote operations](https://www.terraform.io/docs/backends/operations.html) such as `plan` and `apply`. That is, you use Terragrunt to organize your code and keep it DRY, and you can configure it so that when you run `terragrunt apply` on your computer or in a CI server, it runs `terraform apply` in TFC/TFE rather than your own computer. Under the hood, you can create the workspace automatically, or, re-use a workspace that already exists, including variables and env vars already set in it, such as your AWS access keys, so you don't need to have those locally. You'll also be able to see the `apply` output in the web UI (but not trigger it from the web UI) and see the history of all `apply` calls in the web UI.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+For Terragrunt to use TFC as a remote executor, you'll need a `backend` block. The backend is what determines where Terraform reads state and how execution is handled. In this case, you'll configure TFC as a remote in the backend block. The Terragrunt https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#generate:[`generate` block] can generate the block dynamically so you don't need to include one in each module.
+
+Terraform https://www.terraform.io/docs/state/workspaces.html[workspaces] store the state associated with a given backend. In the Terragrunt configuration hierarchy defined above, each module (`sns` and `sqs`) is a separate workspace and must have its own name in the `backend` block. There are several ways to do this with Terragrunt. The snippet below shows one approach: use the account name (defined in `account.hcl`) as a prefix, and use the directory name of the module as the suffix.
+
+The following snippet is placed in `terragrunt.hcl` at the root of the hierarchy.
+
+[source,hcl]
+----
+locals {
+  tfc_hostname     = "app.terraform.io"
```
brikis98, line 1528: NIT: put a comment here mentioning to change this to their custom URL if using TFE.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+
+In the sections below we will cover each step in turn.
+
+===== Setting up
+
+Consider the following Terragrunt directory structure:
+
+[source]
+----
+.
```
brikis98, line 1502: NIT: give the folder a name, such as `infrastructure-live`. Then in the `terragrunt.hcl` code snippets below, you can specify the full path (i.e., `infrastructure-live/terragrunt.hcl` and `infrastructure-live/dev/us-east-1/sns/terragrunt.hcl`) so it's easier for the reader to tell which file they are looking at.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+The first way is to set two environment variables in the TFC workspace: `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. This is perhaps the simplest approach, because all you need to do to set it up is to generate the access keys for an IAM user, assign that user appropriate permissions, and https://learn.hashicorp.com/terraform/tfc/tfc_workspaces#editing-variables[set the environment variables]. The downside is that the credentials are static. You'll need to create each workspace in advance, rather than relying on the implicit workspace creation, and for each workspace, you'll have to set the environment variables manually. If you rotate the keys, you'll have to update them in each workspace.
+
+A second way is to generate a `provider` block in the root `terragrunt.hcl`, and include the static API credentials in the block. For example:
+
+
+[source,hcl]
+----
+locals {
+  region_vars  = read_terragrunt_config(find_in_parent_folders("region.hcl"))
+  access _key  = "<value for access key>"
```
brikis98, line 1614:
```suggestion
  access_key  = "<value for access key>"
```

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+  secret_key  = "<value for secret key>"
+}
+
+generate "aws_provider" {
+  path      = "provider.tf"
+  if_exists = "overwrite"
+  contents = <<-EOF
+provider "aws" {
+  region     = "${local.region_vars.locals.aws_region}"
+  access_key = "${local._access_key}"
```
brikis98, line 1624:
```suggestion
  access_key = "${local.access_key}"
```

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+                       "${local.profile_name}",
+                       "--role-arn",
+                       "${local.role_arn}")))
+  access_key     = local.creds[0]
+  secret_key     = local.creds[1]
+  session_token  = local.creds[2]
+}
+
+generate "aws_provider" {
+  path      = "provider.tf"
```
brikis98, line 1653: There are a few major drawbacks with all these approaches that use `generate` to create a `provider.tf` file:

1. You end up with that generated file on your local disk, with your creds in plaintext.
1. Worse yet, that file is _also_ uploaded to TFC/TFE, and stored there in plaintext.

When you add env vars in the TFC/TFE UI, you can at least mark them as "sensitive," so they are stored in an encrypted format. 

The biggest issue of all, however, is that, AFAIK, 90% of the reason to use TFC/TFE as a remote backend is so that developers do NOT need to have AWS creds of any kind locally. For big companies, distributing such creds is a messy affair, and many of them use TFC/TFE specifically to avoid that.

Therefore, my vote is that we recommend the following:

1. Create your workspaces first. An admin can do this manually... Or, perhaps you set up your Terragrunt folder structure, and run `validate-all`. I _think_ that will run `init` in each module, which will create all the workspaces.
1. Have an admin provide machine user AWS credentials for each workspace. E.g., With TFE, you could assign an IAM role to the server. With TFC, the admin may have to go into each workspace and set the access keys as env vars in the UI, marking them as "sensitive." One other option for admins could be to use the [tfe_workspace_ids data source](https://www.terraform.io/docs/providers/tfe/d/workspace_ids.html) to periodically retrieve all workspaces (set `names` to `*`) and then loop over them using `for_each` in a [tfe_variable resource](https://www.terraform.io/docs/providers/tfe/r/variable.html) to add the AWS access keys as env vars.
1. Once all workspaces are configured, now the devs can run `terragrunt xxx` and it will execute in that workspace with the credentials it needs.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+
+
+===== Summarizing Terragrunt and TFC/TFE compatibility
+
+Using the combination of features above, you can effectively use Terragrunt with TFC or TFE as a remote backend. Terragrunt can generate backend blocks, tfvars files, and provider configurations, resulting in dynamic workspace creation. Commands like `terragrunt apply-all` will operate in the same way as they do without TFC/TFE, calling each module dependency in order, and allowing for passing outputs between modules.
+
+IMPORTANT: The https://gruntwork.io/reference-architecture/[Gruntwork Reference Architecture] is not built for use with TFC or TFE. Although the reference architecture is deployed using Terragrunt, the state is stored in S3, and the configurations discussed above are not included. Migrating from the standard Gruntwork reference architecture to TFC is not supported.
```
brikis98, line 1675: This makes it sound like migrating to TFC is impossible, when in reality, it's just tedious. To do it, you'd have to update the root `terragrunt.hcl` file(s) to use TFC/TFE as a backend, run `init` in each module, and enter "yes" when prompted if you want to migrate the state. Once you do that, you can use the Ref Arch with TFC/TFE as a remote backend (but still can't trigger things from the UI).

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+
+1. In the Workspace settings, in the SSH key section, you can now choose the Gruntwork access key.
+
+.Configuring a TFC Workspace to access Gruntwork repositories
+image::/assets/img/guides/infrastructure-as-code-library/tfc_ssh_key.png[]
+
+
+When TFC runs your module, it will use the configured SSH key to clone the Gruntwork repository during the run, and will continue with creating the SNS topic you defined. You can reuse this SSH key across any other workspaces that depend on Gruntwork modules.
+
+That's it! You can use this pattern for any of Gruntwork's Terraform modules. This even works for modules like github.com/gruntwork-io/terraform-aws-eks[`terraform-aws-eks`] which need to download external dependencies (like https://github.com/gruntwork-io/kubergrunt[`kubergrunt`]). Those modules use the special https://github.com/gruntwork-io/package-terraform-utilities/tree/master/modules/executable-dependency[`executable-dependency`] module to install the external dependency within the executor at run time.
```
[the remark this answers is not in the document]
zackproser: It's not obvious from the diff here but the `terraform-aws-eks` link is broken.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
bwhaley: I was trying to find a balance between an informative section on the topic and full fledged tutorial on using TFC. Sounds like you think the tutorial approach is going to be more helpful. I don't necessarily disagree, but I do wonder whether this guide is really the right place for such a tutorial. Should we make this a standalone guide? It would be relatively short in comparison to the other guides, but then wouldn't constitute a major section within this guide.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
bwhaley: > There are a few major drawbacks with all these approaches that use generate to create a provider.tf file:

At least in the case of the `run_cmd` approach the credentials are temporary. The script could generate credentials that are only valid for 15 minutes. Seems like a reasonable balance to me, but admittedly it isn't a super clean solution. I think we should at least note it as an option.

> The biggest issue of all, however, is that, AFAIK, 90% of the reason to use TFC/TFE as a remote backend is so that developers do NOT need to have AWS creds of any kind locally. For big companies, distributing such creds is a messy affair, and many of them use TFC/TFE specifically to avoid that.

These seems anecdotal to me. It may be one reason, but it may equally be that they want to track the runs centrally, or use TFC/TFE for storing state, or want sentinel... or many other reasons. I honestly can't imagine trying to be a developer trying to do stuff in the cloud without any kind of credentials said cloud. But I suppose from a separation of duties standpoint, a prod environment might be read-only for the developer, in which case the approach described above would be a concern. But then why not just have whomever the admin is run terraform?

I don't see manually creating workspaces first as a practical solution. The workspaces are likely to change often. Seems pretty error prone. But probably worth writing up as another alternative.

Yet another option would be a wrapper for terragrunt! It could request temporary AWS credentials, put them in a place where terraform can reach them (either in the workspace using the `tfe_variable` resource, as you mentioned, or in a `*.auto.tfvars` file, in a terragrunt.hcl... numerous ways to do it), then invoke terragrunt, and finally clean up after itself when complete.

However, this extra layer of abstraction seems a little hacky too.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
bwhaley: My point was that we don't want to support this.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,272 @@ module "vpc" {
...
+==== Using a Gruntwork module in a TFC workspace
+
+One of the first tasks to undertake in TFC is to create a workspace, which is a backend location where Terraform state
+is kept. Workspaces are usually linked to VCS repositories that contain the Terraform code that defines your infrastructure.
+In a typical TFC workspace, after linking your VCS repository, you add Terraform variables to customize your
+environment, then use the built-in plan and apply features to manage the infrastructure. Under the hood, TFC clones the
+repository, sets up the variables according to your definitions, and runs Terraform against the configuration. Each run
+is tracked and available for review, and the state is stored centrally within TFC. 
+
+When you're using Gruntwork modules, you use the source attribute on a module block to read a module from a Gruntwork
```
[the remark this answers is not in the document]
bwhaley: It wasn't meant as an ideal integration, more as an example of how to do _any_ kind of integration. I was only using this as an example. But now I'm thinking that a more full-fledged, standalone guide might be even better. I'll address in a future commit.

bwhaley:
Thanks to everyone for the reviews! There is a lot of good feedback here. After ruminating on this for a bit, I've come to the conclusion that this content could be a standalone guide. To fully flesh this out, we should have a complete example with more screenshots and steps laid out. That's too much content for the existing "How to use the library" guide. I worry that it would make the original guide feel cluttered and a bit confused. On its own, it will be somewhat short relative to some other guides, but will do a better job of explaining exactly what is going on without confusing matters. Furthermore, it might be better for SEO and discoverability. Will mention folks when the new guide is ready for a review.

brikis98:
I agree with the conclusion of not making that single guide even longer, but one concern I have is that pulling out TFC/TFE into a separate guide, while everything else remains in this one guide, gives the impression that TFC/TFE has an oddly high status with us, when the reality is exactly the opposite. 

Should we consider a more general breakdown of the "How to use the Gruntwork Infrastructure as Code Library" into several guides? E.g.,

1. A general overview (the current guide), which then links to a couple other guides for detailed instructions.
1. Terraform. The detailed instructions for vanilla Terraform, Terragrunt, and TFC / TFE.
1. Scripts, binaries, and Helm. The detailed instructions for Packer, Docker, gruntwork-install, etc. 
1. Ref Arch. This will likely be something we do as part of the service catalog work, so no need to do anything with this now.

bwhaley:
Okay, I've rewritten this to incorporate everyone's feedback, and I'd really appreciate a re-review. Although I still think it's pretty long for this already long guide, refactoring this in to multiple guides is more involved than I had an appetite for right now. Let me know what you think of the new and much improved version.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,363 @@ module "vpc" {
...
+  Using this approach, Terraform modules are still defined in an `infrastructure-modules` repository as discussed above.
+  Each module has a dedicated https://www.terraform.io/docs/state/workspaces.html[TFC workspace]. You create the
+  workspace in advance, and then you can run the Terraform modules either from the `terraform` CLI or from the TFC UI.
+  In essence, TFC replaces Terragrunt and `infrastructure-live`. You'll be able to use the full TFC feature set, but you
+  won't be able to take advantage of the benefits provided by Terragrunt.
+
+Use TFC with Terragrunt::
+  Alternatively, you can use both `infrastructure-modules` and `infrastructure-live` repositories as described above,
+  storing the wrapper modules in `infrastructure-modules`, and using `infastructure-live` and Terragrunt for
+  deployments. In this approach, TFC is used a https://www.terraform.io/docs/backends/types/remote.html:[remote backend]
```
yorinasub17, line 1428:
```suggestion
  deployments. In this approach, TFC is used as a https://www.terraform.io/docs/backends/types/remote.html:[remote backend]
```

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,363 @@ module "vpc" {
...
+
+In the sections above, we discussed how to use Gruntwork with two VCS repositories: `infrastructure-modules`, containing your Terraform code that wraps the modules from Gruntwork Library, and `infrastructure-live`, containing Terragrunt configurations that enable you to manage Terraform easily across multiple accounts and environments. When using Gruntwork with TFC, you have two choices regarding these repositories:
+
+Use TFC without Terragrunt::
+  Using this approach, Terraform modules are still defined in an `infrastructure-modules` repository as discussed above.
+  Each module has a dedicated https://www.terraform.io/docs/state/workspaces.html[TFC workspace]. You create the
+  workspace in advance, and then you can run the Terraform modules either from the `terraform` CLI or from the TFC UI.
+  In essence, TFC replaces Terragrunt and `infrastructure-live`. You'll be able to use the full TFC feature set, but you
+  won't be able to take advantage of the benefits provided by Terragrunt.
```
brikis98, line 1423:
```suggestion
  won't be able to take advantage of the benefits provided by Terragrunt (e.g., keeping your code and variables, CLI args, etc DRY and adding support for applying changes across multiple modules / environments).
```

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -1399,6 +1409,363 @@ module "vpc" {
...
+1. The `terraform` block at the top uses the Gruntwork `sqs` module from https://github.com/gruntwork-io/package-messaging/[`package-messaging`].
+1. The `include` block includes the configuration from the parent directories. This is how the remote `backend` block from the root `terragrunt.hcl` is included.
+1. The `locals` block reads the values from `common.hcl` in the root of the hierarchy, making them available for local reference.
+1. Finally, the `generate` block creates a file called `terragrunt.auto.tfvars`. Like the `backend.tf` file, this file will be generated alongside the rest of the `*.tf` files that Terragrunt downloads from the `sqs` module, making those inputs available for TFC to read when running `terraform` commands in the remote executor.
+
+Any of the inputs needed by the module must be included in the generated tfvars file. In the configuration above, only
+the `name` variable is specified. Most modules will need more configuration.
+
+
+===== Summarizing Terragrunt and TFC/TFE compatibility
```
brikis98, line 1761: Add one section just before this where you run `terragrunt apply` and show in the CLI output that the `plan` and `apply` actually happen remotely, in TFC, which stores the state and maintains a log of all these runs.

yorinasub17:
I just realized that we should update `terragrunt.gruntwork.io` to reference this when it launches.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
bwhaley: ed6c196

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
bwhaley: ed6c196

bwhaley:
> I just realized that we should update terragrunt.gruntwork.io to reference this when it launches.
Good thinking. Which section do you think it belongs in? As a top level item in the _Community_ section, perhaps?

yorinasub17:
> Good thinking. Which section do you think it belongs in? As a top level item in the Community section, perhaps?

Hmm I think a top level item in "Feature" is probably the best place, considering that this will be in the same vein as [the docs for aws auth](https://terragrunt.gruntwork.io/docs/features/aws-auth/). The community section is really about the Terragrunt/Gruntwork community.

bwhaley:
> Hmm I think a top level item in "Feature" is probably the best place, considering that this will be in the same vein as [the docs for aws auth](https://terragrunt.gruntwork.io/docs/features/aws-auth/). The community section is really about the Terragrunt/Gruntwork community.

Now I am wondering if this entire terragrunt section really belongs in the Terragrunt documentation rather than in this guide. The guide could just point to those docs instead. What do you think?

yorinasub17:
> Now I am wondering if this entire terragrunt section really belongs in the Terragrunt documentation rather than in this guide. The guide could just point to those docs instead. What do you think?

Now that you mention it, yes. I think that would help for better visibility with the terragrunt community (a couple of folks have asked about TFC support, e.g https://github.com/gruntwork-io/terragrunt/issues/938). That said, I'm not sure it is worth the refactor right now, since you will probably want to also update the examples to be more generic.

We should probably launch as is, file a ticket in `terragrunt`, and then consider that when we do the massive refactor for this guide where we break it up. What do you think?

bwhaley:
> That said, I'm not sure it is worth the refactor right now, since you will probably want to also update the examples to be more generic.
> 
> We should probably launch as is, file a ticket in terragrunt, and then consider that when we do the massive refactor for this guide where we break it up. What do you think?
I'm really on the fence about it. I actually don't think that there would be too much work involved to make it more generic. There are only a couple of references to the gruntwork GitHub repos. More of the work would be in refactoring for markdown... which is also not too big of a deal. But I'm also sensitive to the amount of time already spent, plus I'd like to follow up with RV again.

Since this isn't urgent, let's wait on the merge and release to see what @brikis98 thinks.

brikis98:
My vote is to launch as-is, start getting value out of this immediately (i.e., marketing value from announcement, retention improvement with existing customers, etc), and refactor later.

bwhaley:
Thanks for all the feedback! I'll go ahead and merge this in.

eak12913:
Just deployed this to production. One item to note during the deployment is that I saw a bunch of warnings from the static site ascii doc generator:

```
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1473: list item index: expected 2, got 1
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1475: list item index: expected 3, got 1
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1704: list item index: expected 2, got 1
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1706: list item index: expected 3, got 1
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1746: list item index: expected 2, got 1
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1747: list item index: expected 3, got 1
asciidoctor: WARNING: 2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc: line 1748: list item index: expected 4, got 1
```

I'm investigating this now. Hopefully it's not a problem.

PR #297, opened 2020-05-28
Feature/compliance pricing

oredavids:
- Enables AWS CIS Compliance subscription on the checkout page

May-28-2020 12-08-11


josh-padnick:
Just checked this by running it locally and it looks great!

PR #301, opened 2020-05-29
Add guides to navbar [WWW-27]

oredavids:
Screen Shot 2020-05-29 at 3 20 42 PM

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -72,14 +72,22 @@
                   >Achieve Compliance</a
                 >
               </li>
+              <li>
+                <a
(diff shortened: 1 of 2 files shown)


josh-padnick:
Looks great!

oredavids:
@eak12913 the sole purpose of the Id selector on that item is to be able to target it & separate the center nav items from the nav items on the right. I simply renamed it to appropriately reflect the new nav item.

An outer div could instead be introduced for the center nav items and targeted instead but I didn't want to spend any additional time to optimize the CSS given other priorities.

PR #303, opened 2020-05-30
Chore/restore cookie notification [WWW-24]

oredavids:
Screen Shot 2020-06-01 at 6 16 03 PM

Below are the issues that were identified & resolved.
- The intercom app_id was only being set when the application permits hotjar recording. Fixed by removing the dependency.
- The intercom setup didn't call intercom's boot method as specified here: https://developers.intercom.com/installing-intercom/docs/basic-javascript#section-step-2-launch-messenger
- The cookie notification display was predicated on not only intercom being loaded and but also opened by the user. Fixed by removing the dependency.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -41,46 +41,56 @@
     a.appendChild(r);
   })(window, document, "https://static.hotjar.com/c/hotjar-", ".js?sv=");
 </script>
-<script>
-  window.intercomSettings = {
-    app_id: "rzcjacgd"
-  };
-</script>
 {% endif %}
 
 <!-- Intercom -->
 <script>
-  (function() {
+  console.log("Runnning this script!!!!!")
+  window.intercomSettings = {
+    app_id: "rzcjacgd"
+  };
+</script>
+<script>
+  (function () {
     var w = window;
     var ic = w.Intercom;
     if (typeof ic === "function") {
       ic("reattach_activator");
-      ic("update", intercomSettings);
+      ic("update", w.intercomSettings);
     } else {
       var d = document;
-      var i = function() {
+      var i = function () {
         i.c(arguments);
       };
       i.q = [];
-      i.c = function(args) {
+      i.c = function (args) {
         i.q.push(args);
       };
       w.Intercom = i;
-      function l() {
+      var l = function () {
         var s = d.createElement("script");
         s.type = "text/javascript";
         s.async = true;
-        s.src = "https://widget.intercom.io/widget/rzcjacgd";
+        s.src = `https://widget.intercom.io/widget/${w.intercomSettings.app_id}` ;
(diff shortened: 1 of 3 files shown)


josh-padnick:
I love the Sizzy screenshot! So much information in a single image. Overall, this looks great. The one observation I have is that it looks like we need to vertically center the text. E.g. on the Nexus 7, it renders like this:

image

It's the only screen width that uses this format but occupies just one line of text instead of two. Other than that this is ready to launch.

_includes/scripts.html
```diff
@@ -41,46 +41,55 @@
...
         i.c(arguments);
       };
       i.q = [];
-      i.c = function(args) {
+      i.c = function (args) {
         i.q.push(args);
       };
       w.Intercom = i;
-      function l() {
+      var l = function () {
```
eak12913, line 69: What's `var l` for? It doesn't appear to be used

_includes/scripts.html
```diff
@@ -41,46 +41,55 @@
...
       }
     }
-  })();
+})();
 </script>
+<script>
+  window.Intercom("boot", {
+    app_id: window.intercomSettings.app_id
+  });
+</script>
```
eak12913, line 91: What's this for?

_includes/scripts.html
oredavids: It is later called on line 78, 80 & 82. I copied the latest install script from Intercom.

_includes/scripts.html
oredavids: To start up the intercom messenger. https://developers.intercom.com/installing-intercom/docs/basic-javascript#section-step-2-launch-messenger

oredavids:
@eak12913 I have responded to your questions and added a description of the issues that were noticed and fixed to the PR description.

josh-padnick:
> I'd like to understand what does this cookie pop-up change have to do with Intercom.

If you look at the screencap I posted above, you'll notice the layering order:

1. Cookie popup
1. Intercom icon
1. Website

Long ago, in an earlier iteration of the site, Intercom was on top and blocked the "OK" button, so the insight was that we need to be intentional about their ordering.

oredavids:
@eak12913 please take a look at the responses to your comments.

_includes/scripts.html
eak12913: Ah, I see. Could this have a better name like "bootstrapIntercom" or something along those lines? I always get concerned with single character variable names :-/

oredavids: I'm leaving this as-is simply because this is the default script from Intercom's guides. If we want to change the variable name; we should do that for all the other single character variables contained in the script as well.

oredavids:
@eak12913 kindly deploy when you check in tomorrow.

PR #315, opened 2020-06-09
[WWW-39] Update Favicon

eak12913:
This PR updates the favicon to the on that was used by the V2 site. This replaces the historical purple grunty favicon as our site doesn't really have a purple color palate any more.


oredavids:
@eak12913 I was looking through this card yesterday and I thought it required a bit more clarity. The screenshot was from the repo-browser page which had the new grunty favicon. And it seemed @josh-padnick was stating that the background of the grunty in the favicon needed to be changed. It isn't clear if Josh intended that we change the favicon on the static site in addition to the background change.

eak12913:
Josh and I were pairing when this issue was created. I am pretty sure that his intention is simply for us to us the favicon from the V2 site and use it instead of the purple one the static site uses right now.

eak12913:
Merging this PR now and we can always tweak this Favicon later if we need to.

PR #318, opened 2020-06-10
[WWW-34] Fix AsciiDoctor Warnings

eak12913:
One of our guides (How To Use Gruntwork IaC Lib) had some ordered lists which were written using the markdown style of putting `1.` for each ordered list item. AsciiDoctor has a different syntax for rendering ordered lists which is to just use a `.` (a period, without the "one").

I've updated the AsciiDoc to reflect this syntax change and the warning are now resolved.


oredavids:
@eak12913 ready to merge?

PR #321, opened 2020-06-10
Feature/dpa page [WWW-56]

oredavids:
- Adds Data Processing Agreement Page [WWW-56]

_Page_ style was made to sync with the rest of the legal pages (cookie & privacy policy).

Screen Shot 2020-06-11 at 11 37 54 AM

[View Page here](https://deploy-preview-321--keen-clarke-470db9.netlify.app/legal/dpa/)

[WWW-56]: https://gruntwork.atlassian.net/browse/WWW-56


josh-padnick:
Looks good, Ore. I guess we don't have a link to this yet because we're doing this as part of the larger epic around setting up the legal section? 

Separately, here are some suggestions for improvement:

1. More consistent header. Ideally, you could do something like this:
image

But if that takes a lot of time, it's probably not worth it.

2. You could make the download action more prominent by using a secondary-style button:

image

Note that I took this style from the button at the top of https://deploy-preview-321--keen-clarke-470db9.netlify.app/achieve-compliance/ and used the following for the icon per https://fontawesome.com/v4.7.0/icon/download:

```html
<i class="fa fa-download" aria-hidden="true"></i>
```

Also, can you add the link for the DPA document itself?

oredavids:
@josh-padnick:

> More consistent header. Ideally, you could do something like this:

The consistency I was aiming for was with the rest of the "Legal" Pages (Cookie Policy and Privacy Policy). They also do not have this highlighted header as I referenced in the PR description. Would you like to change their headers as well?

> You could make the download action more prominent by using a secondary-style button:

I will add that.

josh-padnick:
> The consistency I was aiming for was with the rest of the "Legal" Pages (Cookie Policy and Privacy Policy). They also do not have this highlighted header as I referenced in the PR description. Would you like to change their headers as well?

Ah, good point. These are just legal pages so let's aim for whatever's easiest, which is probably leaving out the fancy header. In that case, the one thing to look into is spacing between the header and the content:

image

vs.

image

Ideally, we'd have a standard `margin-bottom` property on all `<h1>` elements, but since it doesn't look like we  have that, you can just use `margin-bottom: 30px;` for the DPA page to keep things consisent.

oredavids:
@tsmit291 I need an approval to get this merged

PR #324, opened 2020-06-15
Feature/rss page [WWW-58]

oredavids:
- Adds legal RSS feed [WWW-58]

Jun-15-2020 10-43-14

[WWW-58]: https://gruntwork.atlassian.net/browse/WWW-58

### DIFF
--- a/legal.rss
+++ b/legal.rss
@@ -1,0 +1,33 @@
+---
+layout: null
+---
+<?xml version="1.0" encoding="UTF-8" ?>
+<rss version="2.0">
+<channel>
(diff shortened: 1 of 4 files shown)


legal.rss
```diff
@@ -0,0 +1,33 @@
+---
```
eak12913, line 1: Do we want to be committing this file or is this the equivalent to committing the terms of service word doc? If we don't want to ever commit this file, we should consider adding it to the `.gitignore`

oredavids: This is the template for what's served in the RSS feed so it needs to be committed. It's separate from the terms-of-service word-doc.

PR #325, opened 2020-06-15
Create new "Trust" page [WWW-59]

oredavids:
- Adds Trust page [WWW-59]

_I raised some questions in the Jira ticket around the wordings of RSS feed notification and where we link to this page on the website._

_Also, the links for the RSS feed, the DPA and Subprocessors page will not be routed correctly in this preview environment because the implementations of those pages are in separate PRs which are yet to be merged._

Screen Shot 2020-06-15 at 12 56 09 PM

[View Page in Preview Env](https://deploy-preview-325--keen-clarke-470db9.netlify.app/trust/)

[WWW-59]: https://gruntwork.atlassian.net/browse/WWW-59

### DIFF
--- a/pages/trust/_legal-docs.html
+++ b/pages/trust/_legal-docs.html
@@ -1,0 +1,25 @@
+<div class="row row-page-header">
+  <div class="col-xs-12">
+    <ul>
+      <li>
+        <a href="/terms">Terms of Service</a>
+      </li>
+      <li>
+        <a href="/legal/cookie-policy">Cookie Policy</a>
+      </li>
+      <li>
+        <a href="/legal/privacy-policy">Privacy Policy</a>
+      </li>
+      <li><a href="/legal/dpa">Data Processing Agreement</a> (see below)</li>
+      <li>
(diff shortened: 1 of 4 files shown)


pages/trust/_legal-docs.html
```diff
@@ -0,0 +1,25 @@
...
+      <li>
+        <a href="/terms">Terms of Service</a>
+      </li>
+      <li>
+        <a href="/legal/cookie-policy">Cookie Policy</a>
+      </li>
+      <li>
+        <a href="/legal/privacy-policy">Privacy Policy</a>
+      </li>
+      <li><a href="/legal/dpa">Data Processing Agreement</a> (see below)</li>
```
eak12913, line 13: I think the `(see below)` was in the document just for us to "see below for the details of the Data Processing Agreement page" - I don't think it needs to appear here. If it does need to be here, then I'm not sure what we're referring to.

oredavids: @tsmit291 ☝️

tsmit291: yeah we don't need the (see below) text.

PR #327, opened 2020-06-18
Fix rss feed entry encoding

oredavids:


oredavids:
@eak12913 I need approval.

oredavids:
@eak12913 apostrophe is a special character in XML. It can be replaced with `&apos;` but I figured it was arguably better, grammatically, to replace the contraction `we've` with `we have`.

PR #332, opened 2020-07-08
Standardize 'apply now' buttons.

josh-padnick:
Before:

image

After:

image

It's a small change, but makes the way to apply for all positions on the page consistent. Note that "Apply Now" for the SE and SRE roles goes to a mailto:careers@gruntwork.io?subject=xxx link.


oredavids:
This just caught my eye. Should we get rid of the SRE abbreviation for the Senior & Principal roles?
Screen Shot 2020-07-08 at 9 43 26 AM

josh-padnick:
Don't feel too strongly on this one, but I suppose I prefer the isomorphism.

PR #333, opened 2020-07-08
Add element id to IAC cta button on home page

oredavids:
Need this minor change to track the impact(actually no-impact) of removing the full stop in the IAC section header on Users clicking the CTA.

### DIFF
--- a/pages/index/_learn-more.html
+++ b/pages/index/_learn-more.html
@@ -27,6 +27,7 @@
     <p>
       <a
         href="/infrastructure-as-code-library/"
+        id="home-browse-infra-as-code-library"
         class="btn btn-primary"
         role="button"
         ga-on="click"


josh-padnick:
I don't understand the name or use of the id and would appreciate more context, but this change is so minor that I'll approve now!

oredavids:
In the base A/B experiment that I am setting up; I tweaked this heading on the Home page: "Get access to the Infrastructure as Code Library." such that one set of users will see the heading without the full-stop.

The experiment hypothesis is that this change shouldn't result in any considerable change in the number of users, across both experiment variations, that click the "Browse Library" CTA, overall sessions & bounce rates.

I need to add an element ID to that CTA so that our Google tag Manager is set up to accurately track the CTA button clicks.

josh-padnick:
Great, thanks for the explanation!

PR #335, opened 2020-07-09
Lighten the domain expertise requirement

josh-padnick:
@brikis98 and I are actively debating whether domain expertise or product sensibility is the more important quality in our new designer hire. We hope to test this in a trial project, but to avoid designers with limited prior experience in our domain from applying, I'm lightening the domain expertise requirement. Note that we still list "software background" as a bonus, so the core message that we're especially interested in those who already understand our domain is still here.


josh-padnick:
This is a minor change, so I'm going to merge and deploy now.

PR #337, opened 2020-07-10
add redshift to website

mcalhoun:
Add Redshift to the website once [this PR](https://github.com/gruntwork-io/module-data-storage/pull/159) has been merged and released.


josh-padnick:
Nice! Didn't realize we had a RedShift module now! This page is slated for an update soon, but the proposed change looks good.

mcalhoun:
> Nice! Didn't realize we had a RedShift module now! This page is slated for an update soon, but the proposed change looks good.

Well, we didn't, but I decided on using Redshift to house our data warehouse of support data, so instead of just doing clickops I built the module...

PR #340, opened 2020-07-20
[WWW-82] Update RSS feed to reflect all historical terms of service updates

klijakub:
Add all entries in the Terms of Service CHANGELOG to the /legal.rss feed. Some guidelines:

Create one <item> for each CHANGELOG entry. ✔ Done

Use the GitHub commit link as the <link>. ✔ Done

You can copy the text verbatim as is, but feel free to fix any typos you find. 
✔ Done

Delete the Terms of Service CHANGELOG page. 
✔ Done We removed the whole changelog folder from page dir.

Update the link at the top of https://gruntwork.io/terms to point to WWW-78: Render RSS feed on site.
✔ Done

Jira Item: [WWW-82]

[WWW-82]: https://gruntwork.atlassian.net/browse/WWW-82

PR #342, opened 2020-07-20
[WWW-83] Update Terms of Service to latest version

klijakub:
1. Update all “legalese” text at gruntwork.io/terms with the contents of Gruntwork Click-Wrap Terms of Service (FINAL 6.12.20) (2).docx (UPDATE: See comment for latest doc version).✔ Done Update all “Plain English” terms using the other attached Word doc. Note that to do the data entry, it may be easier to render a diff between the attached doc and the current terms of service, which are reflected in the gruntwork-terms.docx file and which I just auto-generated and attached. Be sure to commit all terms changes in a single git commit. ✔ Done

2. Add a new RSS entry announcing the change as follows. It’s expected that this will also render in .  Consider using HTML in the <item> per https://developer.mozilla.org/en-US/docs/Archive/RSS/Article/Why_RSS_Content_Module_is_Popular_-_Including_HTML_Contents. 
✔ Done (We didn't use HTML inside RSS because there were problems to render it correctly inside legal/feed page) 
FAQs

Please also update FAQs in two different places:

https://gruntwork.io/pricing/ ✔ Done

https://gruntwork.io/faq ✔ Done

Jira Item: [WWW-83]

[WWW-83]: https://gruntwork.atlassian.net/browse/WWW-83


oredavids:
LGTM. Feed/RSS comments already captured in my review of #340

josh-padnick:
@klijakub Great work here! Reviewing this content was such a bear that I wound up directly implementing my updates. The only items I think you missed were ending each section title with a `.`, but that was an easy to fix. Otherwise, I made a few changes to the original spec, cleaned up some wording, added in section links, and removed the need to auto-generate the word doc.

One note, if you can set your links as `<a href="#some_target">...</a>` versus `<a href="/terms/#some_target">...</a>`, you get a nice scrolling effect per https://stackoverflow.com/a/3163635/2308858. I fixed that as well.

Thank you for all the hard work on this one, and this is now ready for merge!

PR #348, opened 2020-08-04
[WWW-99] Fix the legal feed

eak12913:
This PR addresses the 404 that users are currently getting when they visit https://gruntwork.io/legal/feed. The reason they are getting the 404 is there was a trailing slash missing from the "permalink" path. This minor issue resulted in Jekyll generating a file named "feed.html" within the `/legal` directory rather than creating a path like: `/legal/feed/index.html`.  Our site is configured to only serve up the default file in each directory and was blocking access to a custom file in a directory that already had an `index.html` file present.


josh-padnick:
Nice find, Eug!

PR #349, opened 2020-08-05
Minimal patches to ensure CI/CD guide is correct

yorinasub17:
This is a minimal patch to stop the bleeding in the guide. Previously the guide recommended to use the latest ecs-deploy-runner version, but the new version has many new features that require explaining (like the script config).

Since it will take some time to write the new version up, I am doing a minimal patch here first that locks the version used so that we don't leave people confused when working through the initial guide.


bwhaley:
LGTM! Reviewed here and in the live guide since it has already been merged.

PR #344, opened 2020-07-23
Update the Landing Zone deployment guide to take into account the latest fixes/code in module-security

brikis98:
See https://github.com/gruntwork-io/module-security/releases/tag/v0.34.0 and https://github.com/gruntwork-io/module-security/releases/tag/v0.34.1 for details.

PR #352, opened 2020-08-07
Add note on ulimit

brikis98:
As per https://github.com/gruntwork-io/gruntwork-io.github.io/pull/344#discussion_r466689610, running `apply` on the `account-baseline-xxx` modules may give you the error `Error: pipe: too many open files`. I added a note on increasing `ulimit` to fix this.


bwhaley:
LGTM!

PR #358, opened 2020-08-31
Fix recaptcha icon display [WWW-103]

oredavids:
Due to Google's terms around displaying the Recaptcha Icon which I have referenced in the [issue here](https://gruntwork.atlassian.net/browse/WWW-103), I have chosen to display it in a different portion of the screen; bottom-left.

- The first pass resulted in this:
Screen Shot 2020-08-31 at 1 23 56 PM

The display on large screens is passable but on mobile; it really starts to overlap too much content and clog the display. This got me thinking that we should not be loading the script on all pages on the website anyway.

We should load Recaptcha only where needed and that also has the added benefit of narrowing the traffic which Recaptcha uses to score interactions and should therefore make it more accurate.

- Final Pass (Recaptcha loaded only on Contact page)
Screen Shot 2020-08-31 at 1 05 18 PM

 
Screen Shot 2020-08-31 at 1 32 23 PM

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -90,9 +90,10 @@
   });
 </script>
 
+{% if page.use_recaptcha %}
 <!-- recaptcha v3 by Google -->
-<script src="https://www.google.com/recaptcha/api.js?render={{ site.recaptchaApiKey }}"></script>
-
+<script src="https://www.google.com/recaptcha/api.js?render={{ site.recaptchaApiKey }}&style='bottomleft'"></script>
+{% endif %}
 <!-- Cookie Policy Notification -->
 <script src="{{ site.assets_base_url }}js/cookie.js" async></script>
 {% endif %}
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -92,7 +92,7 @@
 
 {% if page.use_recaptcha %}
 <!-- recaptcha v3 by Google -->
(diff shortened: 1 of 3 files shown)


josh-padnick:
Thanks, Ore! Rather than have a logo on the left side of every single page, what if we follow [I'd like to hide the reCAPTCHA badge. What is allowed?](https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed). 

We could add the requested context in a non-intrusive style to https://gruntwork.io/contact/.

oredavids:
> Thanks, Ore! Rather than have a logo on the left side of every single page, what if we follow I'd like to hide the reCAPTCHA badge. What is allowed?.

The current implementation shows the logo on **ONLY** the Contact Page.

josh-padnick:
Thanks for the clarification; I actually didn't realize it's only on the contact us page. That being said, this is one of the most important pages on our website and I do worry a bit about this user experience:

image

My guess is that hiding the reCAPTCHA badge and adding the requested text in a non-intrusive style to the button of the contact us page would be a better option than the left-hand logo. Thoughts?

josh-padnick:
I'm just going to merge this as is since we can at least get the reCAPTCHA logo off of every page on the website and to just the contact us page. I'll create a separate story to handle https://github.com/gruntwork-io/gruntwork-io.github.io/pull/358#issuecomment-686833059.

josh-padnick (approved):

PR #360, opened 2020-09-10
Remote Senior Product Designer. Position filled!

josh-padnick:


josh-padnick:
This is a trivial change, so I'm going to merge right away.

PR #361, opened 2020-09-16
Remove Kubernetes Helm Server

bwhaley:
This change is inspired by a [discussion on reddit](https://www.reddit.com/r/devops/comments/itdhwr/terragruntterraform_iaac/g5f3drh?utm_source=share&utm_medium=web2x&context=3) where a user noticed this and asked if we had plans to update it. That repo is deprecated now that we've updated to Helm v3, and it might be confusing to users to see the Tiller reference there. I've also moved the Kubernetes Services Helm chart repo up to replace it.

### DIFF
--- a/_data/library.yml
+++ b/_data/library.yml
@@ -128,29 +128,24 @@
       blurb: use Route53 for ingress hostnames
       description: Deploys Helm chart that installs the external-dns application into your cluster to map Ingress hostnames to Route 53 Domain records.
 
-- name: "Kubernetes Helm Server"
+- name: "Kubernetes Services"
   clouds:
     - AWS
     - GCP
     - Azure
   icons:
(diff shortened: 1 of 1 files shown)


yorinasub17 (approved):
Makes sense to me. We need to complete the migration, which includes archiving that repo and replacing it: https://github.com/gruntwork-io/team-product/blob/master/rfcs/helm3-transition.md#drop-helm-2-support

The only reason why we haven't done it yet is because this was going to be done as part of k8s refactor project (which has been continuously delayed)

bwhaley:
Thanks for the review. Given that this isn't currently dependent upon the rest of the migration/deprecation plan, I'll merge this in so that the web site doesn't confuse anyone.

PR #369, opened 2020-10-19
Remove Matt.

josh-padnick:


brikis98 (commented):
You should remove Matt's headshot from the `/assets/img/headshots` folder too.

josh-padnick:
Ah, good call. Removed. Would you mind approving so I can merge?

brikis98 (approved):
LGTM!

PR #374, opened 2020-11-11
Update sales post to reflect improved understanding.

josh-padnick:


josh-padnick:
Merging now so I can post the changes right away, but feedback welcome as always!

PR #377, opened 2020-11-17
Remove legacy code around MS Word Terms of Service.

josh-padnick:
We removed the automatic generation of an MS Word file a while ago, but there was still some cruft laying around. I also add in `--incremental` to our builds to allow for faster updates when working locally.


eak12913 (approved):
Nice! LGTM!

PR #378, opened 2020-11-18
Update sales position based on latest insights.

josh-padnick:
I keep flipping back and forth but more recently I've come to believe that we want to more strongly consider hiring a sales professional who is tech-savvy. This PR opens up that possibility more explicitly.


infraredgirl (approved):
LGTM!

PR #380, opened 2020-11-18
Fix Code rendering on Guides pages

eak12913:
This PR fixes the visual issues on the guides pages where all of the code blocks were rendered improperly.

The root cause of the issue is that Prism is a library used for automatically syntax highlighting code in `<code>` blocks. This functionality was already being used by our Guides pages to render code. The new Landing Zone page also uses this library but wanted to provide a custom theme. The latter page introduction updated the css and unintentionally overrode the styles for Guides.

This PR brings back the notion of a "default" prism theme and then exposes a page variable that a page can set to have a custom prism theme on a per-page basis.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -11,10 +11,21 @@
   rel="stylesheet"
   type="text/css"
 />
+<!-- The prism CSS may be different depending on which page we're looking at-->
+{% if page.prism_css %}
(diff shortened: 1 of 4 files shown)


josh-padnick (approved):
Awesome turnaround time on this!

PR #381, opened 2020-11-20
Remove links to our -public repos

brikis98:
As part of cleaning up the `-public` repos, we need to remove any inbound links to them. This PR updates the IaC Lib page to replace this links with links to our repo browser. As a bonus, I also add repo browser links to all our repos, and fix broken links on a few items.

### DIFF
--- a/jekyll-serve.sh
+++ b/jekyll-serve.sh
@@ -2,5 +2,4 @@
 
 set -e
 
-bundle exec jekyll build --incremental
 bundle exec jekyll serve --livereload --drafts --host 0.0.0.0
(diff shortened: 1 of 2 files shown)


jekyll-serve.sh
```diff
@@ -2,5 +2,4 @@
 
 set -e
 
-bundle exec jekyll build --incremental
```
eak12913, line 5: @josh-padnick added this recently as an optimization. Is this not necessary?

eak12913 (approved):
One small question but otherwise I spot checked the links and was not able to find any broken ones. LGTM!

jekyll-serve.sh
brikis98: AFAIK it works without it...

brikis98:

PR #382, opened 2020-11-23
[WWW-109] Intercom removed

klijakub:
Hello,

1. The intercom chat window in the bottom right of the website is removed. ✅
2. The site loads correctly without any errors that may have been caused by removing Intercom. ✅

Jira Item: [WWW-109]

[WWW-109]: https://gruntwork.atlassian.net/browse/WWW-109

### DIFF
--- a/_data/cookie-policy.yml
+++ b/_data/cookie-policy.yml
@@ -41,8 +41,7 @@
     - title: Customization cookies
       ans: |
         <p>These cookies enable us to provide unique or custom content based on your preferences, which may include special offers or content that you have shown interest in.</p>
-        <p><strong>Who Sets These Cookies:</strong><br />Intercom, Inc. (<a href="https://intercom.com" title="Google, Inc.">https://intercom.com</a>)</p>
-        <p><strong>How To Opt Out of Them:</strong><br />See the section <a href="#opting-out-of-cookies" title="Opting out of cookies">Opting out of cookies</a> > <a href="#third-party-cookies" title="Third-party cookies">Third-party cookies</a>.</p>
+        <p>We do not currently use any customization cookies. If we add any in the future, we will amend this cookie policy.</p>
 
 - question: What about other tracking technologies?
   items:
--- a/_data/legal.yml
+++ b/_data/legal.yml
@@ -112,3 +112,11 @@
   category: privacy-policy
   link: https://github.com/gruntwork-io/gruntwork-io.github.io/commit/433dfbb
   date: 2020-09-29
+
+- guid: 10014
+  title: Remove Intercom from cookie policy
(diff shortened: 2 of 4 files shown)


eak12913 (commented):
I believe there's also a mention of Intercom in `cookie-policy.yml`. There's also a comment mention in `cookies.js` which should be updated in order to avoid confusion later. Thanks!

klijakub:
> I believe there's also a mention of Intercom in `cookie-policy.yml`. There's also a comment mention in `cookies.js` which should be updated in order to avoid confusion later. Thanks!

Thanks for the review! 
Yeah, I saw that, but don't know which part of the cookie policy you want to delete? The whole paragraph? 
image

Cookie comment from `cookie.js` deleted ;)

eak12913 (dismissed):
Thanks @klijakub, I'm going to approve this PR but I'd like @josh-padnick to weigh in on what to do about the mention in the policy. He can make the final call on it but we should move forward in the mean time.

@josh-padnick once you're back from break - please take a look at https://github.com/gruntwork-io/gruntwork-io.github.io/pull/382#issuecomment-732178905 and let @klijakub know how to remove the mention of Intercom from that page.

josh-padnick:
Chiming in a few days later, but let's update the cookie policy as follows:

**Current:**
```markdown
## Customization cookies
These cookies enable us to provide unique or custom content based on your preferences, which may include special offers or content that you have shown interest in.

### Who Sets These Cookies:
Intercom, Inc. (https://intercom.com)

### How To Opt Out of Them:
See the section Opting out of cookies > Third-party cookies.
```

**New:**
```markdown
## Customization cookies
These cookies enable us to provide unique or custom content based on your preferences, which may include special offers or content that you have shown interest in.

We do not currently use any customization cookies. If we add any in the future, we will amend this cookie policy.
```

Finally, we'll need to update https://gruntwork.io/legal/feed to add the following:

```markdown
2020-12-07

cookie-policy

We've updated our cookie policy to remove Intercom, a vendor we previously used to enable real-time chat via our website.
https://github.com/gruntwork-io/gruntwork-io.github.io/commit/433dfbb
```

Note that you should link to a git commit whose sole update is to remove Intercom from the cookie policy. So you'll really need two git commits here:

1. Remove Intercom from cookie policy
2. Update legal feed

At that point, I think we should be all set here!

josh-padnick:
@klijakub Could you tackle the items in my previous comment so we can get this merged?

eak12913 (approved):
Changes LGTM! @josh-padnick to close the loop here, do we need to cancel any subscription with Intercom?

josh-padnick:
Thank you, @klijakub. This looks great! Merging now. Eugene, I'll cancel Intercom next.

PR #390, opened 2021-01-12
Remove unnecessary package installation

eak12913:
We've recently been seeing errors when running our CI builds for this repo where a specific `systemd` debian package was not able to be found. This was causing frustrating build failures and interfering with our production deploys.

See context:
- https://gruntwork-io.slack.com/archives/C76UJFUEL/p1610465016108300
- https://gruntwork-io.slack.com/archives/C76UJFUEL/p1610471096114400

These errors turned out to be associated with installing a package that we don't need. So I have removed it.

The better solution here would be to switch this build to using a custom docker image that's already been pre-built with all necessary dependencies so that we don't have to keep re-installing things each time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ orbs:
 references:
   base_container: &base_container
     docker:
-      - image: circleci/ruby:2.7.1-buster
+      - image: circleci/openjdk:8u265-jdk-buster
 
   ######################################################################################################################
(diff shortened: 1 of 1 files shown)


pete0emerson (approved):
I watched @eak12913 bang his head on this and we worked through it together in a paired session. We don't need gnupg any more, and gnupg was relying on systemd.

LGTM 🚢

PR #393, opened 2021-01-13
Update Nokogiri dependency

eak12913:
This PR updated the dependency for the nokogiri library. This is a library used by the `jekyll-toc` plugin and is not used directly. The reason to update the version of nokogiri is because of a security notification from Dependabot. I don't believe that there's any issues from making this update.


eak12913:
@oredavids Could you please take a look here and approve if you think it looks good? Thanks!

oredavids (approved):
LGTM & I don't see any errors in the console.

PR #394, opened 2021-01-14
Fix example to be list(string) as var declaration

ina-stoyanova:
[As pointed out by a customer](https://gruntwork-io.slack.com/archives/CAM4CM63C/p1610575381003700), these docs had an example that was using the `iam_group_names_ssh_grunt_sudo_users` and `iam_group_names_ssh_grunt_users` as `string` instead of `list(string)`.

This PR fixes that. Thanks to Nate! 👍


brikis98 (approved):
LGTM!

infraredgirl (approved):
:+1:

PR #419, opened 2021-05-11
Expose git commit SHA at /sha

ebeneliason:
Although I've explicitly set the permalink to `/sha`, this actually only gets served up at `/sha.json`. I'm not quite sure why, or whether it's worth trying to put a redirect in place to enable both paths.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,4 +16,5 @@ end
 # https://github.com/jekyll/jekyll-redirect-from/issues/150
 Encoding.default_external = Encoding::UTF_8
 
+# https://github.com/jekyll/github-metadata — used to obtain git sha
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -15,3 +15,5 @@ end
 
 # https://github.com/jekyll/jekyll-redirect-from/issues/150
 Encoding.default_external = Encoding::UTF_8
+
+gem "jekyll-github-metadata", "~> 2.13"
```
eak12913, line 19: I am not as familiar with Jekyll/Ruby libraries - so perhaps it would be good to add a quick comment here to explain what this gem is for?

eak12913 (commented):
One small request but then LGTM

eak12913 (approved):
LGTM!

PR #420, opened 2021-05-13
[APT-324] Add dev portal sign in link to navbar and footer

ebeneliason:
This adds a sign in link to the nav bar and the connect section of
the footer, which lead to the Gruntwork Developer Portal.

The navbar was using a curious combination of negative margins on
specific list items in order to achieve its layout, which made it
challenging to add another item. This change refactors the layout
into right/left navbar sections which float accordingly, and a
container that adjusts for several screen sizes.

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -19,7 +19,7 @@
         >
       </div>
       <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-2">
-        <ul class="nav navbar-nav navbar-right">
+        <ul class="nav navbar-nav navbar-left">
           <li>
             <a
               href="/how-it-works/"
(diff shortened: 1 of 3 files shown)


_includes/navbar.html
```diff
@@ -98,36 +98,46 @@
               </li>
             </ul>
           </li>
-          <li id="guides-nav-item">
+          <li>
```
eak12913, line 101: Was this style not doing anything?

eak12913: Oh - I think I see - it did have custom styling - but is no longer necessary. Please feel free to disregard!

eak12913 (approved):
LGTM! - but just putting the warning for other folks here again - Please don't merge this PR yet!

PR #421, opened 2021-05-13
Correct exposed path for /sha

ebeneliason:
Previously we were attempting to serve this as .json to obtain an
appropriate content-type, but it could only be reached at /sha.json.
Switching to /sha/index.html so it can be reached at /sha as intended.


josh-padnick (approved):
LGTM!

oredavids (approved):

PR #425, opened 2021-05-21
Fix /sha endpoint

eak12913:
This PR hopes to fix the `/sha` endpoint. I suspect that the reason it's currently not working is that in production we won't serve files that aren't called `index.html`. If the trailing slash is missing from the `/sha` permalink definition then Jekyll will generate the file as `/sha.html`.


ebeneliason (approved):
LGTM

PR #428, opened 2021-05-27
Publish article on Migrating from Ref Arch 1.0 to 2.0

rhoboat:
The links will work once we publish this release: https://github.com/gruntwork-io/infrastructure-live-multi-account-acme/releases/tag/untagged-ea942753304ad7eb0824


bwhaley (dismissed):
This LGTM!

rhoboat:
Hm, I don't like these colors. The "tip" is red, but I'd rather it was the peaceful blue.

rhoboat:
Should I make them this kind?

image

bwhaley:
That's probably the easiest fix if you prefer it! I agree that the red is a bit jarring.

ebeneliason:
> Should I make them this kind?
> 
> image

This looks good to me, and the icon layout looks cleaner as well. I think it's appropriate for the relatively mild level of severity the text conveys. That red is definitely aggressive, _especially_ given the vibrating contrast with the near-cyan blue in the text.

rhoboat:
I was going to add some new styles to the `.admonitionblock` classes we have currently, to support `TIP` (like `.admonitionblock.tip` which is something Jekyll does support!), and I was going to add some new colors, like the muted green or yellow for tips. But I went with the blue. (We only currently have blue and red.)

image

Ready for another review @ebeneliason!

rhoboat:
I noticed that the numbering here is wrong. It should increment:
image

rhoboat:
This is annoying. If I try to "indent" the image, it turns into a code block and the image doesn't show up.

rhoboat:
Ah, I got it working. Need to use single `:` rather than `::`.

bwhaley (approved):

rhoboat:
Thanks for the review. I feel pretty good about it, and I know Eben has at least looked at it. Given that it's blocking all the downstream things we want to launch, I'm going to go ahead and merge this now.

PR #433, opened 2021-06-03
[APT-1192] Update privacy policy

klijakub:
[APT-1192](https://gruntwork.atlassian.net/browse/APT-1192)

This PR updates the privacy policy with the latest changes.

image
image

### DIFF
--- a/_data/legal.yml
+++ b/_data/legal.yml
@@ -119,4 +119,10 @@
   category: cookie-policy
   link: https://github.com/gruntwork-io/gruntwork-io.github.io/commit/da536301
   date: 2020-12-07
-  
+
+- guid: 10015
+  title: Update privacy policy to address our new developer portal
+  description: We've released a new developer portal at https://app.gruntwork.io! So we've updated our privacy policy to address any privacy and data implications that come from its release.
+  category: privacy-policy
+  link: https://github.com/gruntwork-io/gruntwork-io.github.io/commit/101df49e
+  date: 2021-06-03
--- a/_data/privacy-policy.yml
+++ b/_data/privacy-policy.yml
@@ -2,7 +2,7 @@
   items:
     - title:
       ans: |
-        <p>Gruntwork, Inc. (“Gruntwork,” “we,” “our” and/or “us”) respects your privacy and is committed to protecting it through our compliance with this privacy policy (the “Privacy Policy”). Gruntwork is a DevOps platform designed to enable software teams to quickly launch production-grade cloud infrastructure. This Privacy Policy describes the types of information we may collect from and about you when you visit this website (the “Website”); use the Gruntwork services, including all associated features and functionalities, websites, user interfaces, and applications (collectively, the “Services”); engage with us through our newsletters, emails, and branded social media accounts; or otherwise interact with us. It also describes our practices for using, maintaining, protecting, and disclosing that information as well as certain legal rights you may have, subject to applicable law, and how you can exercise them.</p>
+        <p>Gruntwork, Inc. (“Gruntwork,” “we,” “our” and/or “us”) respects your privacy and is committed to protecting it through our compliance with this privacy policy (the “Privacy Policy”). Gruntwork is a DevOps platform designed to enable software teams to quickly launch production-grade cloud infrastructure. This Privacy Policy describes the types of information we may collect from and about you when you visit our website located at <a href="https://gruntwork.io/">www.gruntwork.io</a> (the “Site”); use the Gruntwork services, including our cloud-based developer portal located at <a href="https://app.gruntwork.io">app.gruntwork.io</a> (“the Portal”) and any additional Gruntwork-branded features and functionalities, websites, user interfaces, and applications (collectively with the Portal, the “Services”); engage with us through our newsletters, emails, and branded social media accounts; or otherwise interact with us. It also describes our practices for using, maintaining, protecting, and disclosing that information as well as certain legal rights you may have, subject to applicable law, and how you can exercise them.</p>
         <p>For the purposes of this Privacy Policy, “personal data” means information about an identified or identifiable individual, such as their name or email address and includes “personal data” as defined by the European Union’s General Data Protection Regulation (“GDPR”) and “personal information” as defined by the California Consumer Privacy Act (“CCPA”). Some information that is automatically disclosed to us or collected by us through your use of our Services meets this definition.</p>
 
 - question: To what data does this privacy policy apply?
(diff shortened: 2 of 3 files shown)


_data/legal.yml
```diff
@@ -119,4 +119,10 @@
...
   date: 2020-12-07
-  
\ No newline at end of file
+  
+- guid: 10015
+  title: Update privacy policy to address our new developer portal
+  description: We've released a new developer portal at https://app.gruntwork.io! So we've updated our privacy policy to address any privacy and data implications that come from its release.
+  category: privacy-policy
+  link: https://github.com/gruntwork-io/gruntwork-io.github.io/commit/acefb523
+  date: 2021-06-03
```
ebeneliason, line 128: nit: It was missing before, too, but we could add a newline here.

pages/privacy-policy/index.html
```diff
@@ -32,4 +32,4 @@ <h3>{{ step.title }}</h3>
       </div>
     </div>
   </div>
-</div>
+</div>
```
ebeneliason, line 35: Missing newline.

_data/privacy-policy.yml
```diff
@@ -2,7 +2,7 @@
...
+        <p>Gruntwork, Inc. (“Gruntwork,” “we,” “our” and/or “us”) respects your privacy and is committed to protecting it through our compliance with this privacy policy (the “Privacy Policy”). Gruntwork is a DevOps platform designed to enable software teams to quickly launch production-grade cloud infrastructure. This Privacy Policy describes the types of information we may collect from and about you when you visit our website located at <a href="https://gruntwork.io/">www.gruntwork.io</a> (the “Site”); use the Gruntwork services, including our cloud-based developer portal located at <a href="https://app.gruntwork.io">www.app.gruntwork.io</a> (“the Portal”) and any additional Gruntwork-branded features and functionalities, websites, user interfaces, and applications (collectively with the Portal, the “Services”); engage with us through our newsletters, emails, and branded social media accounts; or otherwise interact with us. It also describes our practices for using, maintaining, protecting, and disclosing that information as well as certain legal rights you may have, subject to applicable law, and how you can exercise them.</p>
```
ebeneliason, line 5: The app URL is correct in the href, but wrong in the text: no `www` before `app.gruntwork.io`.

_data/privacy-policy.yml
```diff
@@ -30,19 +30,36 @@
...
         <h3>Information We Collect From Other Sources</h3>
-        <p>We may also collect information from third parties, such as our service providers, when they are legally allowed to share your information with us. For instance, in some circumstances we may receive limited information about your payment card from our payment processor, such as the last four digits of your card, the country of issuance, and the expiration date.</p>
+        <p>We may also collect information from third parties, such as our service providers, when they are legally allowed to share your information with us. For instance, we may collect personal data about you from the following sources: </p>
+        <ul>
+          <li>
+          <u>Payment processors</u> - In general, your payment card information is provided directly to our third-party payment processor and we do not collect or store that information. However, we may receive limited information from our payment processor such as the last four digits of your card, the country of issuance, and the expiration date.
```
ebeneliason, line 48: Recommend using an em dash (&mdash;) here instead (and remaining bullets)
```suggestion
          <u>Payment processors</u> &mdash; In general, your payment card information is provided directly to our third-party payment processor and we do not collect or store that information. However, we may receive limited information from our payment processor such as the last four digits of your card, the country of issuance, and the expiration date. 
```

_data/privacy-policy.yml
```diff
@@ -30,19 +30,36 @@
...
+          </li>
+          <li>
+          <u>Third-party services</u> - We allow customers to enable a variety of third-party services on our Portal. For instance, users can connect their accounts on Slack and Github to their Gruntwork account. Once enabled, the provider of a third-party service may share certain information with us such as the username and email address associated with that user on the third-party service as well as additional information that the provider has chosen to make available to facilitate the integration. You should check the privacy settings and privacy policies of these third-party services to understand what information may be shared with us. 
+          </li>
+          <li>
+          <u>Another individual at your organization</u> - We may collect your personal data from other individual at your organization who may provide us with your business contact information in the course of using our Services.
```
ebeneliason, line 57:
```suggestion
          <u>Another individual at your organization</u> - We may collect your personal data from other individuals at your organization who may provide us with your business contact information in the course of using our Services. 
```

pages/privacy-policy/index.html
```diff
@@ -1,7 +1,7 @@
 ---
 layout: default
 title: Privacy Policy
-modified_date: September 29, 2020
+modified_date: June 03, 2021
```
ebeneliason, line 4: Shouldn't this date change along with the policy itself, and only the new entry to the feed itself happen in this commit? I haven't done this before, though, so we may need confirmation from @josh-padnick.

pages/terms/index.html
```diff
@@ -94,4 +84,4 @@ <h3 class="col-md-7 col-md-offset-5 col-print-12">
       </div>
     </div>
   </div>
-</div>
+</div>
```
ebeneliason, line 87: One more!

pages/privacy-policy/index.html
ebeneliason: I just looked through the history for this file a bit, which seems to confirm my suspicion. We probably want to lump the date change in with the first commit, and then update the follow-up accordingly with just the feed update (and the new hash).

klijakub: @ebeneliason I don't understand. The hash of the commit shows only the privacy policy updates and not updates to unrelated pages.

klijakub: https://github.com/gruntwork-io/gruntwork-io.github.io/pull/433/commits/acefb523d621da17d2d445221b2f071c1c50efa5

ebeneliason: Well, there are multiple problems now, but the initial issue I'm pointing out here is that this commit (58adbc98ccdb6b152989509ee1735c33430945fe) is intended to do nothing other than add the entry to the feed, with a hash of the commit that represents the change to the policy. However, you're updating the modification date here as well. The update to the date should happen in the same commit as the policy change itself, and the follow-up commit should only modify the feed.

The issue is now further complicated by the follow-up commits needed for the policy document itself. You'll need to rebase the changes, squash the policy fixes into the initial policy change commit, and update the date as well. Then you can take the hash of that, update the feed entry accordingly, and then again squash that down so you're left with 2 clean commits.

1. all policy changes + date
2. feed update with hash from 1

ebeneliason (approved):
Looks good to me, thanks!

PR #436, opened 2021-06-05
Update Signup Complete Page for Dev Portal

eak12913:
This PR updates the Signup complete page with new instructions that are relevant to user's experience given that they will now onboard via the new portal.


ebeneliason (approved):
Love it!

PR #450, opened 2021-06-30
Switch out sales role to sales engineer.

josh-padnick:
Originally, I was going to remove our Sales Lead job from the website, but then I realized that we're actually interested in recruiting candidates for sales engineering already, so I doctored the ad and created a new job in workable to link it to. Feedback welcome!


brikis98 (approved):
LGTM!

PR #459, opened 2021-07-21
Fix links on landing pages

brikis98:
I used relative links in the landing page template originally and those don't seem to work properly when deployed. This PR updates those links to be absolute links (starting with `/xxx`) by pre-pending `page.permalink`.


ina-stoyanova (approved):
LGTM

brikis98:
Thanks for the review! Merging and releasing now.

PR #460, opened 2021-07-21
Tweaks on heroku landing page 

ina-stoyanova:
- [x] Re-recorded gifs & uploaded to Imgur
- [x] Fixed borders around each step on the homepage

### DIFF
--- a/_data/heroku-landing-page-tabs-images.yml
+++ b/_data/heroku-landing-page-tabs-images.yml
@@ -1,9 +1,9 @@
 - id: logs
   title: Logs
-  content: https://imgur.com/TKHOmtI.gif
+  content: https://imgur.com/Z3wQO8G.gif
 - id: metrics
   title: Metrics
-  content: https://imgur.com/ceSTiup.gif
+  content: https://imgur.com/Bz5r81T.gif
 - id: alarms
   title: Alarms
-  content: https://imgur.com/HuexmVn.gif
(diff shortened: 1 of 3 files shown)


brikis98 (commented):
Thanks for the PR!

One note: we are currently showing these on the landing page at a resolution of 415x293. The gifs on imgur are larger than that, so the browser will resize them, making them blurry. Usually, if you resize it with proper software, it's much more effective than letting the browser do it. Could you give that a shot?

ina-stoyanova:
> Thanks for the PR!
> 
> One note: we are currently showing these on the landing page at a resolution of 415x293. The gifs on imgur are larger than that, so the browser will resize them, making them blurry. Usually, if you resize it with proper software, it's much more effective than letting the browser do it. Could you give that a shot?

I actually did resize them using proper software, but kept the original resolution, which was 850x600. I will reduce them in dimensions too now.

ina-stoyanova:
Okay, after playing around with a few different tools, I'll have to re-record these. Whatever size I reduce them to, they still come up very blurry. And the compression removes extra frames in between, so it looks even worse.

brikis98 (approved):
LGTM!

ina-stoyanova:
Thanks for the review! Will merge and release now!

brikis98:
Don't forget to do a new release to deploy these changes!

PR #463, opened 2021-07-21
Rename hipaa-cloud-compliance to hipaa-compliance

pete0emerson:
I just did a quick search and replace. We'll see in the preview if that's sufficient.


bwhaley (approved):
Looks like it works!

PR #464, opened 2021-07-21
Alternative display for included features

bwhaley:
This is meant as an alternative option to the current table display in the HIPAA table since that doesn't render well on mobile.


josh-padnick:
Haven't reviewed the visual updates, but I'd like to share the URL https://gruntwork.io/hipaa-cloud-compliance/ right away with a customer and not have it break when this merges. Could you add the `redirect_from` URL to enable visits to https://gruntwork.io/hipaa-cloud-compliance/ to redirect to https://gruntwork.io/hipaa-compliance/? See https://github.com/gruntwork-io/gruntwork-io.github.io/blob/master/pages/privacy-policy/index.html#L6 for an example. To test the example, visit https://gruntwork.io/privacy-policy and note that it redirects to https://gruntwork.io/legal/privacy-policy.

bwhaley:
The URL rename has been merged [here](https://github.com/gruntwork-io/gruntwork-io.github.io/pull/463) and is being released as we speak, so no need for the redirect.

brikis98 (commented):
Is the core change proposed in this PR:

1. Putting each item in a separate box rather than having them all in one table?
1. Or is it listing each specific HIPAA requirement (e.g., Requirement §164.308(a)(1)(ii)(A))?

I'm very much in favor of (2)! As for (1), I don't have a strong opinion, but the table approach seems a little more coherent / scannable. Also, how much do we care about mobile users here? I would _guess_ most of our users who are going to sign up are developers on desktops, but perhaps a check of GA would help.

bwhaley:
> As for (1), I don't have a strong opinion, but the table approach seems a little more coherent / scannable.

Seems to be the consensus that nobody has a strong opinion, _and_ that the table seems more scannable. BUT, the table isn't very legible on mobile, which is what this is trying to address. I will update this to take in to account Zack's suggestion of using the table on desktop and these boxes on mobile.

> Also, how much do we care about mobile users here? I would guess most of our users who are going to sign up are developers on desktops, but perhaps a check of GA would help.

IMO it's pretty important. People will definitely be seeing this on mobile, esp for their first glance. I'd argue we should _always_ support mobile. If we tweet out a link, for example, people are pretty likely to be seeing it on the go. If it doesn't render well on mobile it's a bad look.

pete0emerson (dismissed):
LGTM

bwhaley:
Pete, thanks for the review, but may I have another? Pushed a small fix.

pete0emerson (approved):

bwhaley:
Thanks! Will release this one now.

PR #487, opened 2021-07-22
Added preface for the features table.

### LINKED ISSUE #471: HIPAA landing page: Preface on features available now vs later
We used to have a preface before going into the table of features available now / later. I think that preface is important. Without it, it’s not clear what we’re looking at. We need to convey in a header + sentence or two that we already have most of the HIPAA stuff there, that you’re starting on a platform that’s already CIS compliant and trusted by thousands of devs, and that we’re actively working on the rest of the HIPAA requirements.

pete0emerson:
This closes #471.

### DIFF
--- a/pages/landing/hipaa/template/index.html
+++ b/pages/landing/hipaa/template/index.html
@@ -79,6 +79,16 @@ footer:
   </div>
   <div class="section">
     <div class="container">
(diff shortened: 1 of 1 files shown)


zackproser (approved):
LGTM

PR #488, opened 2021-07-22
Fix the Gruntwork HIPAA cost in the Why use Gruntwork table

### LINKED ISSUE #481: HIPAA why gruntwork:  monthly cost in table wrong
Monthly cost in this table is wrong.

pete0emerson:
This closes #481.


zackproser (approved):
LGTM

PR #489, opened 2021-07-22
Add hubspot traffic analytics

josh-padnick:
This PR installs the [HubSpot Tracking Code](https://knowledge.hubspot.com/reports/install-the-hubspot-tracking-code) so that we can (a) collect submissions from our contact us forms into HubSpot without using FormBucket, and (b) analyze our site with the [HubSpot Traffic Analytics tool](https://knowledge.hubspot.com/reports/analyze-your-site-traffic-with-the-traffic-analytics-tool). The HubSpot Traffic Analytics tool is an alternative to Google Analytics which may allow us to directly connect website visits to lead submissions and ultimately deals that closed. This will also enable us to "bucket" users who submit forms based on which form they came from.

### What this commit does not do
- It does not remove Google Analytics
- It does not remove FormBucket, though after this merges, I'd like to do that
- It does not change the configuration of any current contact forms, including the new landing pages

### Other notes
Adding this tracker does represent an additional data subprocessor we are using to track customers, so this PR also updates the relevant policies.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -8,6 +8,9 @@
 <script src="/assets/js/{{ js_file }}.js" type="text/javascript"></script>
 {% endfor %} {% endif %} {% if site.gtm_id %}
 
+<!-- HubSpot Embed Code per https://knowledge.hubspot.com/reports/install-the-hubspot-tracking-code -->
+<script type="text/javascript" id="hs-script-loader" async defer src="//js.hs-scripts.com/8376079.js"></script>
+
 <!-- Google Tag Manager (noscript) -->
 <noscript>
   <iframe
--- a/README.md
+++ b/README.md
@@ -40,8 +40,7 @@ To deploy the site:
 1. Hosted on Amazon S3, with CloudFront as a CDN. Using [s3_website](https://github.com/laurilehmijoki/s3_website) to
    automatically upload static content to S3.
 1. We use [Bootstrap](http://www.getbootstrap.com/) and [Less](http://lesscss.org/).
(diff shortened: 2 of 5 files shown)


josh-padnick:
@brikis98 Would you mind approving this? This will give us reports like the following:

image

Note that _traffic source_ can be connected to _new contacts_ (new leads) and _customers_ (who actually signed). One head's up is that all forms on the website will now automatically be submitted to HubSpot as a [non-HubSpot form](https://knowledge.hubspot.com/forms/use-non-hubspot-forms), unless you use Javascript to make clicking the "Submit" button a no-op. This shouldn't affect any existing setup, but we may be able to cancel FormBucket.

_includes/scripts.html
```diff
@@ -8,6 +8,9 @@
 <script src="/assets/js/{{ js_file }}.js" type="text/javascript"></script>
 {% endfor %} {% endif %} {% if site.gtm_id %}
 
+<!-- HubSpot Embed Code per https://knowledge.hubspot.com/reports/install-the-hubspot-tracking-code -->
+<script type="text/javascript" id="hs-script-loader" async defer src="//js.hs-scripts.com/8376079.js"></script>
```
brikis98, line 12: Out of curiosity, is 8376079 our customer ID? Or does HubSpot figure it out from the domain?

brikis98 (approved):
LGTM!

_includes/scripts.html
josh-padnick: Yeah, that's our hubspot customer ID. It's actually included in all our Hubspot URLs.

PR #492, opened 2021-07-22
Added early adoption features on the pricing page.

### LINKED ISSUE #476: HIPAA pricing page: Not clear which benefits are "exclusive"
It isn’t clear which benefits are “exclusive” to the early access customers. We may need a table that shows “early access” in one column and “after launch” in another. Then the early access column can highlight: (a) the lower pricing, (b) perhaps no up-front fee, (c) early access to the code, (d) higher support tier from Gruntwork, (e) audit assistance.

pete0emerson:
This closes #476.

### DIFF
--- a/pages/landing/hipaa/template/pricing/_price-col.html
+++ b/pages/landing/hipaa/template/pricing/_price-col.html
@@ -4,6 +4,23 @@
       <h3 class="h2 no-anchor margin-top-none" style="margin-bottom: 8px;">{{ include.pricing.title }}</h3>
       <p class="description" id="aws-plan">{{ include.pricing.description }}</p>
     </div>
+<hr>
+
+    <div class="row justify-content-center">
+      <h3 class="text-center">Exclusive Benefits for Early Subscribers</h3>
+      <ul class="list-group text-left">
+        {% for feature in include.pricing.early_adoption_features %}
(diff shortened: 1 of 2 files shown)


ebeneliason:
I'm not sure this approach works very well. It takes up so much space that it pushes the price all the way down below the fold, and it also repeats a number of items already shown just below the price. I think it would be more clear to break the list under the price into two sections, which distinguish the exclusive early access perks from the overall offering all in one place.

ebeneliason:
Something like this? (I have code to support this, if it looks like a reasonable direction.)
image

brikis98:
> Something like this? (I have code to support this, if it looks like a reasonable direction.)
> image

I like the two columns... I think the pricing should be in two columns as well. The left one shows the full price, the right shows early access. Oh, and the early access column could be wider / bigger, and the subscriber one a bit grayed out, to indicate (a) normal subscriber isn't available yet and (b) early access is the "best" option 😁

brikis98:
I think this PR is superseded by https://github.com/gruntwork-io/gruntwork-io.github.io/pull/496? Also, see https://github.com/gruntwork-io/gruntwork-io.github.io/pull/496#pullrequestreview-715718913 for my latest thinking on how we can display the early access vs standard more clearly.

PR #493, opened 2021-07-22
Hipaa folder org

ebeneliason:
Removes extra "template" directory from path and creates subdirectories for subpages.


pete0emerson (approved):
Site still works. LGTM.

PR #496, opened 2021-07-22
HIPAA pricing CTA tweak

ebeneliason:
### TODO
- [ ] New Grunty image for Early Access on pricing page 
- [x] Tweak pricing page header text 
- [x] Finish implementing pricing details

### DIFF
--- a/pages/landing/hipaa/pricing/_price-col.html
+++ b/pages/landing/hipaa/pricing/_price-col.html
@@ -12,12 +12,12 @@
         <p class="big-price"><span>$</span>{{ include.pricing.price }} / mo<sup>*</sup></p>
         <p class="medium-price">+ $</span>{{ include.pricing.flat_fee }} flat fee</p>
       </div>
-      <div data-mh="plans-pricing-bottom">
+      <div data-mh="plans-pricing-bottom" style="margin-top: 2em;">
         <p>
           <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance-on-aws/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">Join the waitlist</a>
         </p>
         <p class="text-muted small">
-          <em>Annual contract<br>Billed monthly or annually.</em><br>* All prices are subject to change.
+          <em>* Annual contract, billed monthly or annually.<br>All prices are subject to change. No payment is needed to sign up.</em>
         </p>
       </div>
     </div>
--- a/pages/landing/hipaa/pricing/_price-rows.html
+++ b/pages/landing/hipaa/pricing/_price-rows.html
@@ -15,10 +15,23 @@
     <p class="h1 enterprise-h1">Contact Us</p>
     {% elsif pricing.price == "Free" %}
     <p class="h1 enterprise-h1">Free</p>
-    {% else %}
+    {% elsif pricing.disabled_price %}
+    <p class="big-price-disabled"><span>$</span>{{ pricing.disabled_price }}</p>
+    <p class="help-block-disabled" style="margin-top: -9px">/ month</p>
+    {% if pricing.disabled_flat_price %}
+    <p class="medium-price-disabled">
+      <span>+ $</span>{{ pricing.disabled_flat_price }}
+      <span class="help-block-disabled">flat fee</span>
+    </p>
+    {% endif %} {% else %}
     <p class="big-price"><span>$</span>{{ pricing.price }}</p>
-    <p class="help-block" style="margin-top: -9px">per month</p>
-    {% endif %}
+    <p class="help-block" style="margin-top: -9px">/ month</p>
+    {% if pricing.flat_price %}
+    <p class="medium-price">
+      <span>+ $</span>{{ pricing.flat_price }}
+      <span class="help-block">flat fee</span>
+    </p>
+    {% endif %}{% endif %}
   </td>
   {% endfor %}
 </tr>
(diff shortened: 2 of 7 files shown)


bwhaley (approved):
LGTM!

pages/landing/hipaa/pricing/_price-col.html
```diff
@@ -12,21 +12,22 @@ <h2>Special early access pricing</h2>
...
       </div>
-      <div data-mh="plans-pricing-bottom">
+      <div data-mh="plans-pricing-bottom" style="margin-top: 2em;">
         <p>
-          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">I'm interested</a>
+          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">Join the waitlist</a>
         </p>
         <p class="text-muted small">
-          <em>Annual contract<br>Billed monthly or annually.</em><br>* All prices are subject to change.
+          <em>* Annual contract, billed monthly or annually.<br>All prices are subject to change. No payment is needed to sign up.</em>
```
brikis98, line 20: A few thoughts:

1. "All prices are subject to change" seems to go against the, "Lock in a reduced rate for your subscription, in perpetuity!" benefit for early access subscribers. Might want to remove it.
1. Perhaps instead of "No payment is needed to sign up" we say something like "No payment is needed to join the waitlist."

brikis98 (commented):
A few high level thoughts on the pricing page:

1. We're missing all the info on user limits (e.g., max 20) and per-user pricing. The FAQ mentions it, but not the pricing page itself.
1. The structure of the page still feels like it's not quite telling the story of "you're signing up for early access to this full product that's in development" as well as it could. See below for a proposed alternative, which addresses this issue and the previous one.

screencapture-gruntwork-io-migrate-from-heroku-to-aws-pricing-2021-07-27-11_41_25

pages/landing/hipaa/pricing/index.html
```diff
@@ -24,9 +24,18 @@
...
         description: Deploy dev and stage infrastructure today and start building your application. HIPAA compliance coming soon!
+      - name: Discount pricing
+        description: Lock in a reduced rate for your subscription, in perpetuity!
+      - name: No up-front fee
+        description: Skip the up front cost completely.
+      - name: Direct support
+        description: Get a higher level of support from our team of experts at no extra cost.
+      - name: HIPAA audit assistance
+        description: We’ll work with you throughout your audit process and make any security changes required.
+    features:
```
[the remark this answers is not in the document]
ebeneliason: I agree that audit assistance should, at least for now, only be an early access feature. Also, I believe we weren't allowed to mention HITRUST specifically (@bwhaley can you confirm?), so we should change the phrasing to something like "HIPAA compliance audit assistance."

pages/landing/hipaa/pricing/_price-col.html
ebeneliason: 1. These two clauses are trying to say different things, but I agree they are unclear. The former is intended to make clear that they aren't locking in the prices shown on this page, which may yet change. The early access feature is a discounted rate lock in, just at a rate that may not be reflected here. How about "The price shown is not yet final"?
2. That's much clearer. Let's go with your suggestion.

ebeneliason:
I like the suggested page layout from Jim, above. @zackproser I might not have time to get into this for a bit. If you have the time do you want to take a stab at the proposed layout?

zackproser:
> I like the suggested page layout from Jim, above. @zackproser I might not have time to get into this for a bit. If you have the time do you want to take a stab at the proposed layout?

Yup - I will take care of it :+1:

pages/landing/hipaa/pricing/_price-col.html
brikis98: Ah, I see what you're driving at. Well, I _think_ for anyone that joins the waitlist at price X, we probably should honor price X and not change it on them... Otherwise, that really is a bait & switch. Perhaps we should even include the price as a hidden element in the form they submit so we have that data. That way, if we change it for later experiments, it'll only affect later visitors.

So do we really need the "price shown is not yet final" piece?

zackproser:
> > I like the suggested page layout from Jim, above. @zackproser I might not have time to get into this for a bit. If you have the time do you want to take a stab at the proposed layout?
> 
> Yup - I will take care of it +1

Basic table put in place in 2cdda2d8. Now working on tuning the pricing presentation to match the mockup.

ebeneliason:
In discussion with @zackproser, we agreed we should adjust the copy at the top of the page to add clarity to what's being offered. A proposal:

**Gruntwork HIPAA Early Access Program**
Sign up now for early access to our HIPAA-compliant infrastructure and reference applications! Early access subscribers will receive exclusive benefits and a reduced subscription price.

pages/landing/hipaa/pricing/index.html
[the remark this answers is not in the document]
zackproser: Yes, Ben said we should not use HITRUST but instead "HIPAA-compliant `<thing>`, so this seems correct. It's now implemented in the pricing table, too, so I'm going to mark this as resolved.

pages/landing/hipaa/pricing/_price-rows.html
```diff
@@ -0,0 +1,127 @@
...
+      <span>+ $</span>{{ pricing.disabled_flat_price }}
+      <span class="help-block-disabled">flat fee</span>
+    </p>
+    {% endif %} {% else %}
+    <p class="big-price"><span>$</span>{{ pricing.price }}</p>
+    <p class="help-block" style="margin-top: -9px">/ month</p>
+    {% if pricing.flat_price %}
+    <p class="medium-price">
+      <span>+ $</span>{{ pricing.flat_price }}
+      <span class="help-block">flat fee</span>
```
ebeneliason, line 32: This class includes `display: block`, which is forcing the span onto the next line. Not sure what that class was intended for, but it looks like the only style rule that actually matters here is `font-size: 16px`.

ebeneliason: Follow up thought. Would it be better to just advertise "no flat fee" instead? I'm not sure myself — just presenting the option. 
image

zackproser: Ah - thanks! Fixed in 6b347422

ebeneliason:
This looks great overall. I know there's a few little things to tweak including the new Grunty image, but I'm going to approve and merge this so we can build atop it, and so future previews can include everything we have so far, combined.

ebeneliason:
This looks great overall. I know there's a few little things to tweak including the new Grunty image, but I'd like to approve and merge this so we can build atop it, and so future previews can include everything we have so far, combined.

Unfortunately, for some reason this includes a change from me and so I can't approve it.

rhoboat (approved):
Looks good to Eben! I haven't taken a close comb through, myself.

PR #502, opened 2021-07-23
New image for HIPAA always-up-to-date section

### LINKED ISSUE #470: HIPAA landing page: Step 5 screenshot changes
In Step 5, “Stay up to date automatically via Terraform pull requests,” the current screenshot is a bit hard to read. Perhaps instead, we show a screenshot of versioned releases from the terraform-aws-eks repo?. Or a screenshot from a RenovateBot PR showing the diff page with an upgrade from version X to Y?

ebeneliason:
Closes #470


zackproser (approved):
LGTM

brikis98 (commented):
Thanks for the PR Eben! I think one issue is that the image becomes too blurry to read when shrunk down to the 500px space available for it on the page. If you resize the image explicitly to 500px, rather than having the browser do it, it's a tiny bit better, but still not readable.

Here's a zoomed in screenshot that is 500px wide that I think is a little more legible:

Screen Shot 2021-07-27 at 12 05 28 PM

ebeneliason:
Although larger, this is also quite blurry, partly because it's not high-DPI compatible. The right fix is probably twofold:
1. Zoom in the browser _before_ taking the screenshot, so that the font size is rendered larger with respect to the page layout.
2. Capture an image at 2x the intended display resolution (e.g. 1000px if we want to display it at 500px wide) so that it has double the sharpness on high density displays (most, these days), which should keep the text legible.

I'll see if I can reproduce one like that you've shown here.

Issue #474, opened 2021-07-22
HIPAA technical details page: add app and infra "Details" sections


zackproser:
The zig zag portion is a bit short. See [this comment](https://github.com/gruntwork-io/gruntwork-io.github.io/pull/456#discussion_r673986479) for ideas on a few more zigs/zags and more info to put below that.

ebeneliason:
Related — we should probably fix the height of the code window and have it scroll as needed so that a) it doesn't cause re-layout when changing tabs and b) it doesn't push the rest of the page content way down. Maybe a fixed height of `450px` or so?

The one gotcha I see here is that the little red label from the figure is inside the element which will need to scroll, so we may need to hide it, shift it fully inside its container (which will need `overflow: scroll`), or use another element in its parent instead.

PR #503, opened 2021-07-23
Fix code viewer height - replace red badge with comment

zackproser:
These changes fix the issues raised in [this comment](https://github.com/gruntwork-io/gruntwork-io.github.io/issues/474#issuecomment-885281283) on the application templates (technical details) page. Namely, they: 
* Replace the code viewer's red label with code comments that say the same thing
* Prevent the code viewer from causing re-layout on tab navigation by constraining its maximum height to 450px
* Add `overflow-y: scroll` to the code viewer so that examples longer than 450px automatically scroll.

### DIFF
--- a/_data/hipaa-landing-page-tabs-code.yml
+++ b/_data/hipaa-landing-page-tabs-code.yml
@@ -1,9 +1,13 @@
 - id: runapp
   title: Bash
   content: |
-    <figure><figcaption>Read Secrets From AWS Secrets Manager on startup</figcaption>
+    <figure>
       <pre>
       <code class="language-bash">
+      ######################################
+      # How to read from AWS Secrets Manager 
+      # ####################################
+       
       # Returns the raw value of a secret string stored in AWS Secrets Manager.
       function read_secret_from_aws_secrets_manager {
         local -r environment_name="$1"
@@ -27,9 +31,12 @@
 - id: database
   title: database.js
   content: |
-    <figure><figcaption>Use database secrets to connect securely</figcaption>
+    <figure>
       <pre>
       <code class="language-javascript">
+        /****************************************************
+        * How to use config to connect to a database securely
+        *****************************************************/
         const { sendBackendResponse, sendResponse } = require("../views/views.jsx");
         const db = require("../models/db");
         const config = require("config");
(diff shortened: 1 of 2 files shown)


_data/hipaa-landing-page-tabs-code.yml
```diff
@@ -27,9 +31,12 @@
 - id: database
   title: database.js
   content: |
-    <figure><figcaption>Use database secrets to connect securely</figcaption>
+    <figure>
       <pre>
       <code class="language-javascript">
+        /****************************************************
+        * How to use config to connect to a database securely
+        *****************************************************/
```
ebeneliason, line 39: Nit: should there be a newline after these comment blocks just like the first above?

pages/landing/hipaa/technical-details/_how-it-works.html
```diff
@@ -32,7 +32,7 @@
             </div>
           {% endfor %}
         </div>
-        <div class="tab_code_content">
+        <div style="max-height: 450px; overflow-y: scroll;"class="tab_code_content">
```
ebeneliason, line 35: Nit: Should we just set this as fixed instead of max, so that the container size doesn't change at all?

ebeneliason (dismissed):
LGTM overall — left a couple of nits.

pages/landing/hipaa/technical-details/_how-it-works.html
zackproser: Yes - fixed in 67ed3b7

_data/hipaa-landing-page-tabs-code.yml
zackproser: Good catch! Added in 67ed3b7

ebeneliason (approved):
Looks good.

I'll note two other little UI nits I noticed, but let's not hold this up. I played with this myself yesterday and wasn't able to easily resolve…
1. The blue border around the tabs isn't showing up for some reason. But, I see the CSS for it…
2. The gray background on the code block that's shorter than the code box should ideally extend to fill the box. Maybe the background color is on the wrong element.

zackproser:
> Looks good.
> 
> I'll note two other little UI nits I noticed, but let's not hold this up. I played with this myself yesterday and wasn't able to easily resolve…
> 
>     1. The blue border around the tabs isn't showing up for some reason. But, I see the CSS for it…
> 
>     2. The gray background on the code block that's shorter than the code box should ideally extend to fill the box. Maybe the background color is on the wrong element.

Roger that - thanks. I will move these to a follow-up issue. 

Thanks for the review. Going to merge this in now.

zackproser:
Captured the follow-ups in https://github.com/gruntwork-io/gruntwork-io.github.io/issues/504

PR #506, opened 2021-07-23
HIPAA Landing Page: Fix contact button, redirect from /hipaa-compliance.

josh-padnick:
Previously clicking the "I'm interested" button went to a 404. This PR fixes that.

In addition, there is now a redirect from `/hipaa-compliance` to `/hipaa-compliance-on-aws`.

### DIFF
--- a/pages/landing/hipaa/pricing/_price-col.html
+++ b/pages/landing/hipaa/pricing/_price-col.html
@@ -14,7 +14,7 @@
       </div>
       <div data-mh="plans-pricing-bottom">
         <p>
-          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance-on-aws/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">I'm interested</a>
+          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance-on-aws/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">Join the waitlist</a>
         </p>
         <p class="text-muted small">
           <em>Annual contract<br>Billed monthly or annually.</em><br>* All prices are subject to change.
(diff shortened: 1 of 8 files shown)


pages/landing/hipaa/pricing/_price-col.html
```diff
@@ -14,7 +14,7 @@ <h2>Special early access pricing</h2>
       </div>
       <div data-mh="plans-pricing-bottom">
         <p>
-          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">I'm interested</a>
+          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance-on-aws/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">I'm interested</a>
```
ebeneliason, line 17:
```suggestion
          <a id="pricing-cta-aws" class="btn btn-info pricing-ctas" href="/hipaa-compliance-on-aws/contact" ga-on="click" ga-event-category="{{ page.path | slugify }}" ga-event-action="select-cta">Join the waitlist</a>
```

In the thread we talked about changing this to a stronger CTA, like the suggestion, or even just "Request Early Access". It looks like that hasn't actually happened yet. Do you want to sneak it in here?

ebeneliason (approved):
Looks good to me. I wish there were a better solution for relative paths in links to avoid breaking things if the permslinks change, but this looks correct as it is now.

pages/landing/hipaa/pricing/_price-col.html
josh-padnick: Sure, "request early access" sounds great, so I'll go with that.

josh-padnick: Actually, we've already got that as a link on the homepage, so I'll go with your suggestion.

PR #507, opened 2021-07-23
Attempt to fix redirect issue in prod.

josh-padnick:
The HIPAA thank you page works fine locally, but not in prod. I'm hoping this trailing `/` is the issue.


josh-padnick:
Update: This seemed to fix the issue! TL;DR: A trailing `/` is required for `permalink` properties to work properly in GitHub Pages.

PR #508, opened 2021-07-26
Update IaC Library guide to use Service Catalog

rhoboat:
- [x] Remove `infrastructure-modules` from main instructions.
- [ ] What do we do about section on Deploying using Plain-Old-Terraform?
- [ ] How should the TFE/TFC section be modified to mention service catalog instead of `infrastructure-modules`?

### DIFF
--- a/_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
+++ b/_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
@@ -238,11 +238,11 @@ IMPORTANT: You must be a [js-subscribe-cta]#Gruntwork subscriber# to access thes
   so you know how to run things in dev, how to deploy changes to prod, how to find metrics and logs, how to connect
   over VPN and SSH, and so on. This is a great starting point for exploring the Reference Architecture.
 +
-* https://github.com/gruntwork-io/infrastructure-modules-multi-account-acme[infrastructure-modules]: In this repo,
+* https://github.com/gruntwork-io/terraform-aws-service-catalog[AWS Service Catalog]: In this repo,
   you'll find the reusable modules that define the infrastructure for the entire company (in this case, for Acme).
   These are like the blueprints for a house.
 * https://github.com/gruntwork-io/infrastructure-live-multi-account-acme[infrastructure-live]: This repo uses
-  the modules from `infrastructure-modules` to deploy all of the live environments for the company (dev, stage, prod,
+  the modules from `terraform-aws-service-catalog` to deploy all of the live environments for the company (dev, stage, prod,
   etc). These are like the real houses built from the blueprints.
 * https://github.com/gruntwork-io/sample-app-frontend-multi-account-acme[sample-app-frontend]: This repo contains a
   sample app that demonstrates best practices for a Docker-based frontend app or microservice, including examples of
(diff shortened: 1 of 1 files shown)


_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -238,11 +238,11 @@ IMPORTANT: You must be a [js-subscribe-cta]#Gruntwork subscriber# to access thes
   so you know how to run things in dev, how to deploy changes to prod, how to find metrics and logs, how to connect
   over VPN and SSH, and so on. This is a great starting point for exploring the Reference Architecture.
 +
-* https://github.com/gruntwork-io/infrastructure-modules-multi-account-acme[infrastructure-modules]: In this repo,
+* https://github.com/gruntwork-io/terraform-aws-service-catalog[AWS Service Catalog]: In this repo,
   you'll find the reusable modules that define the infrastructure for the entire company (in this case, for Acme).
```
yorinasub17, line 242: This sentence needs rewording, as it is no longer code for a particular company.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -238,11 +238,11 @@ IMPORTANT: You must be a [js-subscribe-cta]#Gruntwork subscriber# to access thes
...
   These are like the blueprints for a house.
 * https://github.com/gruntwork-io/infrastructure-live-multi-account-acme[infrastructure-live]: This repo uses
-  the modules from `infrastructure-modules` to deploy all of the live environments for the company (dev, stage, prod,
+  the modules from `terraform-aws-service-catalog` to deploy all of the live environments for the company (dev, stage, prod,
   etc). These are like the real houses built from the blueprints.
```
yorinasub17, line 246: This should point to `for-production` examples in the service catalog.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -640,140 +640,124 @@ next two sections of the guide will walk you through how to use each of these.
...
+        └ variables.tf
 ----
 
-==== Configure your providers
+==== Terraform and providers are configured
 
-Inside of `main.tf`, configure whatever Terraform providers you're using. Since the `vpc-app` module you're using in
-this guide is an AWS module, you'll need to configure the AWS provider:
+The service catalog module's `main.tf` file configures the Terraform providers to use. Since the `vpc-app` module in
+this guide is an AWS module, it configures the AWS provider in the configuration of Terraform itself in `main.tf`:
```
yorinasub17, line 676: This is actually not true. It only specifies required provider versions, but it does not actually configure an `aws` provider. If you follow this as is, you will actually run into a prompt asking to specify the AWS Region to deploy into.

The old docs are still accurate even for the service catalog flavor.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -640,140 +640,124 @@ next two sections of the guide will walk you through how to use each of these.
...
-  link:/guides/foundations/how-to-configure-production-grade-aws-account-structure[How to Configure a Production Grade AWS Account Structure]).
+Pin the Terraform version::
+  The code above will ONLY allow you to run it with a specific Terraform version. This is a safety measure to ensure
+  you don't accidentally pick up a new version of Terraform until you're ready. This is important because (a) Terraform
+  is a pre 1.0.0 tool, so even patch version number bumps (e.g., `0.12.6` -> `0.12.7`) are sometimes backwards
+  incompatible or buggy and (b) once you've upgraded to a newer version, Terraform will no longer allow you to deploy
+  that code with any older version. For example, if a single person on your team upgrades to `0.12.7` and runs `apply`,
+  then you'll no longer be able to use the state file with `0.12.6`, and you'll be forced to upgrade everyone on your
+  team and all your CI servers to `0.12.7`. It's best to do this explicitly, rather than accidentally, so we recommend
+  pinning Terraform versions.
```
yorinasub17, line 708: This part is incorrect, as the service catalog module has a relaxed terraform version `>= 0.12.26` and not a strict `=` setup.

The advice is relevant though, and I would recommend having version locking in the root module calling the service catalog.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -640,140 +640,124 @@ next two sections of the guide will walk you through how to use each of these.
...
-  backend "s3" {}
-
-  # Only allow this Terraform version. Note that if you upgrade to a newer version, Terraform won't allow you to use an
-  # older version, so when you upgrade, you should upgrade everyone on your team and your CI servers all at once.
-  required_version = "= 0.12.6"
+variable "cidr_block" {
+  description = "The IP address range of the VPC in CIDR notation. A prefix of /18 is recommended. Do not use a prefix higher than /27. Examples include '10.100.0.0/18', '10.200.0.0/18', etc."
+  type        = string
 }
-----
```
yorinasub17, line 737: NOTE: This section is still relevant for calling service catalog module with pure terraform. The Service Catalog modules do not configure a backend.

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -640,140 +640,124 @@ next two sections of the guide will walk you through how to use each of these.
...
-  is a pre 1.0.0 tool, so even patch version number bumps (e.g., `0.12.6` -> `0.12.7`) are sometimes backwards
-  incompatible or buggy and (b) once you've upgraded to a newer version, Terraform will no longer allow you to deploy
-  that code with any older version. For example, if a single person on your team upgrades to `0.12.7` and runs `apply`,
-  then you'll no longer be able to use the state file with `0.12.6`, and you'll be forced to upgrade everyone on your
-  team and all your CI servers to `0.12.7`. It's best to do this explicitly, rather than accidentally, so we recommend
-  pinning Terraform versions.
+----
 
-==== Use the modules from the Gruntwork Infrastructure as Code Library
+==== Gruntwork Infrastructure as Code Library modules are referenced
```
yorinasub17, line 741: The changes to this section is a bit confusing to me. Specifically, the user doesn't (and shouldn't) need to modify the `module` blocks in the service catalog module to modify the vars.

In fact, the references to library modules are useful for informational purposes, but is actually irrelevant when it comes to using the service catalog module as a user in pure terraform, since it's embedded within the service module (and hence the user can't change it without forking).

_posts/2019-08-26-how-to-use-gruntwork-infrastructure-as-code-library.adoc
```diff
@@ -797,261 +781,115 @@ Versioned URL::
...
-
-When you're done testing, clean up by running:
-
-[source,bash]
-----
-terraform destroy ../
-----
 
 [[automated_tests_terraform]]
 ==== Automated tests for Terraform code
```
yorinasub17, line 793: Like above, this section is a little bit confusing. It is useful as informational purposes, but is only relevant to users who would be building their own module, or will be contributing code.

Perhaps we can streamline this section to focus entirely on what a consumer of the service catalog would need to do, and separate the contribution question to another part of the docs?

yorinasub17 (commented):
The terragrunt section updates LGTM, but the terraform updates feel confusing:

- I think the previous content is still relevant, for users who don't want to/can't use the service catalog module. Those users need to create their own service catalog module, and this content is accurate for that use case. Perhaps we can keep it as is, but move it to the bottom of the doc as an advanced use case?

- The new content about using the service catalog module is incomplete. Take a look at https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/core-concepts.md#deploy-new-infrastructure to see what is required to deploy the service catalog module. Also see comments for specific parts that are confusing.

- Also at a high level, a lot of the content around the service catalog is/(is going to be) duplicated with [service catalog docs](https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/core-concepts.md#deploy-new-infrastructure). Perhaps we should merge those docs here and then have the service catalog docs point to here so we only have one place of docs to keep up to date?

rhoboat:
Thanks for comments, Yori. I wanted to start the conversation, as it was easier to make the confusing edits than it was to write up why it's so confusing. But you hit on exactly what I was facing here, with a few additions.

> I think the previous content is still relevant, for users who don't want to/can't use the service catalog module. Those users need to create their own service catalog module, and this content is accurate for that use case (see docs on creating your own service catalog). Perhaps we can keep it as is, but move it to the bottom of the doc as an advanced use case?

I think it could be argued that _this_ article is about how to use the IAC library modules, not the service catalog. The "M" in MSA (module catalog). So it could be argued that we scrap these edits, keep this article the same, with all the same mentions of `infrastructure-modules`.

> The new content about using the service catalog module is incomplete. Take a look at https://github.com/gruntwork-io/terraform-aws-service-catalog/blob/master/core-concepts.md#deploy-new-infrastructure to see what is required to deploy the service catalog module. Also see comments for specific parts that are confusing.

Thanks for pointing out these docs, I didn't know they existed, and I'm sure a lot of folks don't. Is it worth having a guide on the website that does little more than link to this content?

> Also at a high level, a lot of the content around the service catalog is/(is going to be) duplicated with service catalog docs. Perhaps we should merge those docs here and then have the service catalog docs point to here so we only have one place of docs to keep up to date?

I agree about having a single place. Not sure where that single place should be. It works fine in my view for it to be the docs in GitHub, but give it some "real estate" representation on the website somehow.

yorinasub17:
> Thanks for pointing out these docs, I didn't know they existed, and I'm sure a lot of folks don't. Is it worth having a guide on the website that does little more than link to this content?
> I agree about having a single place. Not sure where that single place should be. It works fine in my view for it to be the docs in GitHub, but give it some "real estate" representation on the website somehow.

Yea I'm not sure what the best place is either... Both docs actually are not very discoverable. The one in the service catalog repo is not discoverable because I don't think a lot of people really go through the root README, and the one on the gruntwork.io website is just too damn long and doesn't have a search bar.

I think if it were up to me, I would go for a dedicated docs site that is optimized for "how to use gruntwork," and each of these would be there, similar to terragrunt.gruntwork.io and terratest.gruntwork.io. But that is probably too expensive. Perhaps the short term is to create dedicated entries in the GitHub discussions forum that links to each of these articles if and when we roll that out?

ebeneliason:
Hey @rhoboat, we're about to merge the change which will redirect all guides links to docs.gruntwork.io, which will remove this file entirely and invalidate this PR. Are these changes still required, and if so can they be made against the Markdown in the docs repo?

josh-padnick:
Should we merge this or close this?

eak12913:
Closing this PR as stale and no longer relevant.

PR #517, opened 2021-07-29
Remove left padding so logo aligns with body copy

ebeneliason:
This will affect all landing pages, but the change is trivial, and just fixes the logo alignment.


brikis98 (approved):
LGTM

Issue #515, opened 2021-07-29
Pricing table doesn't work well on mobile


ebeneliason:
On mobile the table doesn't fit and create an awkward horizontal scrolling situation. I'm not sure there's a cheap solution here — we could try to add the class so it's at least contained within the page boundary, but that doesn't make it more legible; just less broken…
<img width="731" alt="image" src="https://user-images.githubusercontent.com/687448/127571749-c8d0a847-7724-4229-b846-47d8eb7b96d5.png">

zackproser:
Yeah, I noticed this as well yesterday :(

zackproser:
Worth looking into if the Heroku page suffers from the same issue.

zackproser:
Thanks @ina-stoyanova and @marinalimeira for volunteering to dig into this!

zackproser:
@ebeneliason Did you want to keep this one open? Or do we consider the latest PR that adds horizontal scrolling to be sufficient?

ina-stoyanova:
@zackproser & @ebeneliason I'd suggest closing this one off, so we get a cleaner view of the Issues section :) I'll wait a bit and close it off today!

zackproser:
Thanks @ina-stoyanova! I think we're good to close this one.

PR #519, opened 2021-07-30
Fix for broken horizontal scroll

ina-stoyanova:
Fixing this issue: https://github.com/gruntwork-io/gruntwork-io.github.io/issues/515

- this is fixing the horizontal scroll on the parts where that's not expected - like the header.


ina-stoyanova:
The text appears a bit too wide - there's a media query from the `.table-responsive` class. Left it as it is for now, but will review again if it looks too bad.

zackproser (approved):
LGTM! Thanks for digging into this.

ina-stoyanova:
Thanks for the approval - merging now!

PR #525, opened 2021-08-01
Fix reference-infra link + tabbed nav

### LINKED ISSUE #520: Technical details nav links don't work for HIPAA page
On the HIPAA page, if you go to "Technical Details" in the nav, there are two problems:

1. The text "Reference Infrastructure" should actually be "Reference Architecture."
1. When you click the "Reference Infrastructure" link, you are taken to the wrong tab of the technical details page. The URL is https://gruntwork.io/hipaa-compliance-on-aws/technical-details#reference-infra, which looks right, but the tab that's visible is "Reference Applications" instead of "Reference Architecture."

zackproser:
Fixes #520 

Without the trailing slash, the guard in `main.js` that scopes tabbed navigation to the HIPAA technical details page will prevent the navigation code from running.

### DIFF
--- a/_includes/landing-navbar-hipaa.html
+++ b/_includes/landing-navbar-hipaa.html
@@ -67,7 +67,7 @@
                     ga-on="click"
                     ga-event-category="nav-{{ page.path | slugify }}"
                     ga-event-action="hipaa-reference-architecture"
-                    >Reference Infrastructure</a
+                    >Reference Architecture</a
                   >
                 </li>
                 <li>
--- a/_includes/landing-navbar-hipaa.html
+++ b/_includes/landing-navbar-hipaa.html
@@ -63,7 +63,7 @@
               <ul class="dropdown-menu dropdown-menu-left" role="menu">
                 <li>
                   <a
-                    href="{{ landing_page_base_url }}/technical-details#reference-infra"
+                    href="{{ landing_page_base_url }}/technical-details/#reference-infra"
                     ga-on="click"
                     ga-event-category="nav-{{ page.path | slugify }}"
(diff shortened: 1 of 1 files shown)


_includes/landing-navbar-hipaa.html
```diff
@@ -63,7 +63,7 @@
               <ul class="dropdown-menu dropdown-menu-left" role="menu">
                 <li>
                   <a
-                    href="{{ landing_page_base_url }}/technical-details#reference-infra"
+                    href="{{ landing_page_base_url }}/technical-details/#reference-infra"
```
brikis98, line 66: Thanks for the quick fix Zack! Note there's one more piece to this fix: updating the text from "Reference Infrastructure" to "Reference Architecture" a few lines below.

zackproser: Thanks, Jim :+1: Note that the use of "Infrastructure" there was actually intentional, since we were not sure that newcomers to the page would understand what an "Architecture" is in the context of Gruntwork, whereas we figured "Infrastructure" was generic enough to be generally understood at first glance, hopefully. 

That said, I'm happy to switch it back, too.

zackproser: Updated in dc174ee

brikis98: I don't feel too strongly about it, but the page itself has the title "Reference Architecture," and that's what we use elsewhere, so perhaps consistency is the better route?

brikis98 (approved):
LGTM!

zackproser:
Thanks for the review! Going to merge this in now.

PR #556, opened 2021-08-05
Install LinkedIn Insights / ads tracking tag

### LINKED ISSUE #555: Install LinkedIn Insights tracking tag
For our landing page experiments, our SEM expert needs the LinkedIn tracking tag installed globally across our site.

zackproser:
Fixes #555

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -50,6 +50,26 @@ css styling -->
 </script>
 <!-- End Google Tag Manager -->
 
+<!-- LinkedIn Insights Ad tag -->
+<script type="text/javascript">
+_linkedin_partner_id = "3333610";
+window._linkedin_data_partner_ids = window._linkedin_data_partner_ids || [];
+window._linkedin_data_partner_ids.push(_linkedin_partner_id);
+</script><script type="text/javascript">
+(function(l) {
+if (!l){window.lintrk = function(a,b){window.lintrk.q.push([a,b])};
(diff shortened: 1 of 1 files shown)


eak12913 (approved):
LGTM! I'm seeing this tracking being blocked by default by my browser (Brave) and similarly it's being blocked by Safari. Not sure if this will cause data issues later as other browsers may not block tracking by default like this.

zackproser:
Thanks for the review, Eugene! Going to merge this in now so we can release and verify it.

josh-padnick:
Just a head's up guys that adding LinkedIn means we're now setting third-party cookies for our users, so we have to update our Cookie Policy and our legal changelog. See https://github.com/gruntwork-io/gruntwork-io.github.io/pull/489 for an example of what was needed for this when I added HubSpot. If we're just doing this temporarily, it's not necessary to add, but if we're going to keep it permanent, we'll need to update the Cookie Policy and changelog

zackproser:
> Just a head's up guys that adding LinkedIn means we're now setting third-party cookies for our users, so we have to update our Cookie Policy and our legal changelog. See #489 for an example of what was needed for this when I added HubSpot. If we're just doing this temporarily, it's not necessary to add, but if we're going to keep it permanent, we'll need to update the Cookie Policy and changelog

@josh-padnick Roger that. Thanks for the heads up. I will set a reminder for a week to re-evaluate where we are here.

PR #579, opened 2021-08-10
Update pricing table for eks-for-enterprise

yorinasub17:
This implements the suggestion from https://gruntwork.atlassian.net/browse/WWW-119 to switch over the pricing model from EKS addon to being included in the Enterprise plan (as opposed to being a separate add-on to the subscription like CIS).

PR #582, opened 2021-08-11
Follow up to EKS enterprise pricing table

yorinasub17:
Address comments in https://github.com/gruntwork-io/gruntwork-io.github.io/pull/579

### DIFF
--- a/pages/landing/enterprise-eks/pricing/index.html
+++ b/pages/landing/enterprise-eks/pricing/index.html
@@ -19,34 +19,34 @@ categories:
       - name: Core services
         description: Suite of core Kubernetes applications every EKS cluster should have, including AWS LoadBalancer controller, external-dns, cluster-autoscaler, and more.
         plans: ["Standard", "Enterprise"]
-      - name: Basic CI/CD
+      - name: Standard CI/CD
         description: Support for setting up an infrastructure CI/CD pipeline using traditional CI services (e.g., Jenkins, CircleCI, GitLab CI, etc).
         plans: ["Standard", "Enterprise"]
-      - name: Basic monitoring
+      - name: Standard monitoring
         description: Support for metric and log aggregation to AWS CloudWatch.
         plans: ["Standard", "Enterprise"]
-      - name: Basic deployment models
+      - name: Standard deployment models
         description: Support for rolling deployments where a configurable number of service Pods are replaced at a time.
         plans: ["Standard", "Enterprise"]
-      - name: Basic networking
+      - name: Standard networking
         description: Support for AWS VPC based networking for Kubernetes Pods.
         plans: ["Standard", "Enterprise"]
-      - name: Basic auto scaling
+      - name: Standard auto scaling
         description: Support for EKS worker node auto scaling throught the Kubernetes cluster-autoscaler.
         plans: ["Standard", "Enterprise"]
-      - name: Advanced CI/CD
+      - name: Enterprise CI/CD
         description: Support for CI/CD workflows using Kubernetes native services such as Argo CD and Flux.
         plans: ["Enterprise"]
-      - name: Advanced monitoring
+      - name: Enterprise monitoring
         description: Support for metric and log aggregation through various 3rd party and in-cluster options, such as DataDog, Prometheus, Grafana, Elastic Stack, etc.
         plans: ["Enterprise"]
-      - name: Advanced deployment models
+      - name: Enterprise deployment models
         description: Support for wide range of deployment models including canary, blue/green, automatic rollbacks, etc.
         plans: ["Enterprise"]
-      - name: Advanced networking
+      - name: Enterprise networking
         description: Support for wide range of Pod networking options including Calico and Service Mesh.
         plans: ["Enterprise"]
-      - name: Advanced auto scaling
+      - name: Enterprise auto scaling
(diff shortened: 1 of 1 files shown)


pages/landing/enterprise-eks/pricing/index.html
```diff
@@ -19,7 +19,7 @@
       - name: Core services
         description: Suite of core Kubernetes applications every EKS cluster should have, including AWS LoadBalancer controller, external-dns, cluster-autoscaler, and more.
         plans: ["Standard", "Enterprise"]
-      - name: CI/CD
+      - name: Basic CI/CD
```
yorinasub17, line 22: I switched this back based on https://github.com/gruntwork-io/gruntwork-io.github.io/pull/579#discussion_r686719064, but to explain my original reasoning for changing this:

I felt weird calling this "Basic CI/CD" because in some sense this is more advanced as it includes a full workflow engine. The k8s-native solutions don't include workflow engines, so you can't do things like "run test on code and then deploy" - it goes straight to deploy.

What do you think about changing this to "Traditional CI/CD" and "K8s native CI/CD"?

pages/landing/enterprise-eks/pricing/index.html
```diff
@@ -31,10 +31,10 @@
       - name: Basic networking
         description: Support for AWS VPC based networking for Kubernetes Pods.
         plans: ["Standard", "Enterprise"]
-      - name: Cluster auto scaling
+      - name: Basic auto scaling
```
yorinasub17, line 34: Similar to above, I switched this based on https://github.com/gruntwork-io/gruntwork-io.github.io/pull/579#discussion_r686719602 but again I felt a bit weird calling this "Basic" because it's fairly advanced (most basic auto scaling is CPU based, but this is based on Pod scheduling). It also doesn't include pod auto scaling at all, and I was worried calling this "Basic auto scaling" implies that - I have a hunch most people will expect pod auto scaling before cluster auto scaling.

pages/landing/enterprise-eks/pricing/index.html
brikis98: Does Argo CD not work with, say, [Argo Workflows](https://argoproj.github.io/argo-workflows/)?

pages/landing/enterprise-eks/pricing/index.html
brikis98: Well, the implication here is "basic" as in "basic subscription" and not as in "simple functionality." But I suppose there is a little ambiguity there.

Perhaps we can make that explicit? What if we replace the word "basic" with "standard" and the word "advanced" with "enterprise": e.g., "Standard Monitoring" and "Enterprise Monitoring" or "Standard Auto Scaling" and "Enterprise Auto Scaling"?  So now it 100% reflects the subscription types, and hopefully removes ambiguity with how complicated the feature is?

pages/landing/enterprise-eks/pricing/index.html
yorinasub17: They work, but their really two very different systems. I'm not quite sure I follow what you are trying to get at here.

pages/landing/enterprise-eks/pricing/index.html
yorinasub17: I like that, so let's go with it.

yorinasub17: Done: 7df5161

pages/landing/enterprise-eks/pricing/index.html
yorinasub17: Standard / Enterprise seems fine to me, so let's just keep it that way.

brikis98: > They work, but they're really two very different systems. I'm not quite sure I follow what you are trying to get at here.

You wrote:

> The k8s-native solutions don't include workflow engines, so you can't do things like "run test on code and then deploy" - it goes straight to deploy.

So I was wondering if you could implement a similar workflow using Argo Workflows, and trigger it on commits / PRs / etc using Argo CD. In other words, is what you were calling "Traditional CI / CD" really "more advanced" because it supports a "full workflow engine" or can the same things be done with Argo Workflows and therefore, K8S-native CI / CD? Not critical to resolve now, more of a curiosity thing.

brikis98 (approved):
LGTM!

pages/landing/enterprise-eks/pricing/index.html
yorinasub17: Ah gotcha. I was really going for this

> In other words, is what you were calling "Traditional CI / CD" really "more advanced" because it supports a "full workflow engine"

Argo Workflows does indeed give you the same things on Kubernetes, but so does Jenkins-X, or Jenkins with Kubernetes Runners plugins.

What I really wanted to differentiate was between a first class Kubernetes CI/CD experience with self-service and deployment dashboards (which is what Argo CD and Flux gives you), and one that only gives you progress information at the pipeline level, including having to define your custom pipeline in whatever format the pipeline engine requires.

I didn't know what was the right term to use for that, but basic vs advanced felt wrong. Standard vs Enterprise feels a bit better.

brikis98: Roger!

yorinasub17:
Thanks for review. Merging now.

PR #584, opened 2021-08-13
Illustrate How it Works section from Patcher

### LINKED ISSUE #542: Replace Grunty on cubes for the zig zag on the Patcher page
Put in screenshots or `.gif` or more useful images to show what Patcher does.

marinalimeira:
Closes #542.

Preview: https://deploy-preview-584--keen-clarke-470db9.netlify.app/patcher

Some other sugestions:

patcher-1
patcher-2-small
Screen Shot 2021-08-13 at 15 41 30


brikis98:
> patcher-1
> patcher-2-small

Both of these are terrific! I have a slight preference for the second one, as it's smaller, so it fits the page better, and is easier to read / scan.

> Screen Shot 2021-08-13 at 15 41 30

This is great too! The "Patcher discovers dependencies" section could have two tabs, one with this "Web UI" option and one with the "CLI" option above.

brikis98:
Had a few thoughts after looking at this closely. These are NITs, so totally OK to skip if low on time:
> patcher-2-small



1. Might be nice to include some non Gruntwork dependencies in the list. E.g., `express.js` or `rails` or `guava.jar`.
1
> 
> Both of these are terrific! I have a slight preference for the second one, as it's smaller, so it fits the page better, and is easier to read / scan.
> 
> > Screen Shot 2021-08-13 at 15 41 30
> 
> This is great too! The "Patcher discovers dependencies" section could have two tabs, one with this "Web UI" option and one with the "CLI" option above.

ina-stoyanova:
These all look just great! That's some awesome gifs! +1 on Jim's thoughts that the second gif with the smaller list of dependencies is easier to see, but either one is actually quite nice! Maybe the first one is even more realistic...

On the screenshot - that's awesome too! The only feedback there - I don't know what is the "Open Pull request" text. Is that  button that's clickable or status? Maybe it could be clearer.

brikis98:
Had a few more thoughts after looking at the `.gifs` more closely. They are all NITs, so OK to skip or do later.

> patcher-2-small

Might be a good idea to have a mix of Gruntwork and non Gruntwork dependencies. E.g., Perhaps the list is something like:

- `express.js`
- `com.google.guava` 
- `openssl`
- `activerecord`

> Screen Shot 2021-08-13 at 15 41 30

- Same note here about having not only Gruntwork dependencies.
- Current version should probably be different for each rows (not all at `v0.53.1`).
- "Open Pull Request" should probably be in the blue hyperlink color.
- Some of the rows could have a "refresh" icon (to indicate "in progress") and instead of "Open Pull Request," a link to an existing PR (e.g., "Pull Request #344" hyperlink).

marinalimeira:
Thank you Jim and Ina for reviewing this!

> Some of the rows could have a "refresh" icon (to indicate "in progress") and instead of "Open Pull Request," a link to an existing PR (e.g., "Pull Request Update the Landing Zone deployment guide to take into account the latest fixes/code in module-security #344" hyperlink).

Not sure if I follow, does the one below makes sense?

Screen Shot 2021-08-17 at 14 06 39

> Might be a good idea to have a mix of Gruntwork and non Gruntwork dependencies. E.g., Perhaps the list is something like:

Mixing different programming languages is confusing for me. I don't know how to present this information.


Thoughts from a live conversation with Ina:

We will only release patchers for the important dependencies, e.g. React and Rails but not the `react-html-email` dependencies. For JS developers, they can use a [npm update](https://docs.npmjs.com/cli/v7/commands/npm-update) and then open a PR updating _all_ their dependencies, so having a button to update each dependency separately seems counter-intuitive.

marinalimeira:
@brikis98 made some improvements! Final version ready to review: https://deploy-preview-584--keen-clarke-470db9.netlify.app/patcher/

_data/patcher-landing-page-tabs-dependencies-list.yml
```diff
@@ -0,0 +1,7 @@
+- id: cli
+  title: Use the CLI
+  content: /assets/img/landing-page-patcher/patcher-cli.gif
```
brikis98, line 3: I think using a tool like https://asciinema.org/ will allow you to have a purely text based recording of the terminal, rather than an image that is a tiny bit blurry and hard to read. This can be done in a follow-up issue / PR.

_data/patcher-landing-page-tabs-dependencies-list.yml
```diff
@@ -0,0 +1,7 @@
...
+
+- id: ui
+  title: Use the UI
+  content: /assets/img/landing-page-patcher/patcher-ui-tf.png
```
brikis98, line 7: The image here is a little bit blurry. This seems like a minor issue, but as I discussed [here](https://gruntwork.atlassian.net/browse/WWW-124), in my experience, users absolutely "judge a book by its cover," and blurry images may be seen as unprofessional, uncomfortable, etc (see The Iceberg Secret). 

See [here](https://gruntwork.atlassian.net/browse/WWW-124?focusedCommentId=13843) for ideas on how to make the images not blurry. I think a crisp `.gif` of scrolling through the dependencies would be better. This can be done in a follow-up issue / PR.

brikis98: Oh, also, the UI tab should be the first/default one!

pages/landing/patcher/index.html
```diff
@@ -42,33 +44,55 @@
...
+    tab_block: site.data.patcher-landing-page-tabs-dependencies-list
     image_side: left
     body: |
       Patcher will automatically discover dependencies for popular languages and tools such as JavaScript, Ruby, Java, Python, Go, Terraform, and Docker. You can also add special comments to your code to help Patcher discover dependencies in non-standard locations (e.g., in a Bash script).
 
   - heading: Maintainers release new versions and patches
-    image: /assets/img/infrastructure-cube-icons/infrastructure-expert-support.png
-    image_width: 418px
+    precode: |
+      <figure><figcaption>terraform-to-12.patch</figcaption>
```
brikis98, line 61: NIT: on the Use Cases page, we have a `tf-14-to-15.patch` example. We should probably use that same example in both places.

pages/landing/patcher/index.html
```diff
@@ -42,33 +44,55 @@
...
+        <pre>
+          <code class="language-yml">
+      code_transformations:
```
brikis98, line 64: Love the code snippet for this 👍 

NIT: add a comment at the top that says, "This is an example Patch file to update from Terraform XXX to YYY"

pages/landing/patcher/index.html
```diff
@@ -42,33 +44,55 @@
...
+        # Set the required_version parameter in TF { ... } blocks to 0.12
+        - update_required_version_to_tf_12
+        # Update all Gruntwork modules to versions that work with TF 0.12
+        - update_module_versions
```
brikis98, line 68: NIT: make these two comments a tiny bit shorter so no horizontal scrollbar is needed.

pages/landing/patcher/index.html
```diff
@@ -42,33 +44,55 @@
...
+          </code>
+        </pre>
+      </figure>
     image_side: right
     body: |
       As the maintainer of a dependency, you release new versions as usual (e.g., via a new Git tag), but if you are making a breaking change, you can now also create <em>patch files</em> to automatically update your users' code to handle the backwards incompatibilities.
 
   - heading: Patcher automatically patches your code and opens a pull request
-    image: /assets/img/infrastructure-cube-icons/infrastructure-design-your-stack-grunty.png
+    image: /assets/img/landing-page-patcher/terraform-12-pr.png
```
brikis98, line 89: Same comment here on the blurry image. See above for suggestions. One option is to re-use the image here: https://gruntwork.io/eks-for-enterprise#step-7-stay-up-to-date

pages/landing/patcher/index.html
```diff
@@ -42,33 +44,55 @@
...
     image_side: left
     body: |
       Patcher automatically updates you to the new version and runs the maintainer's patches to update your code in case of backwards incompatible changes. All patches are run in a secure, sandboxed environment, so your code is always kept safe.
 
   - heading: You review and merge
-    image: /assets/img/infrastructure-cube-icons/infrastructure-expert-support.png
+    image: /assets/img/landing-page-patcher/review-pr.png
```
brikis98, line 95: This image is great! Note that it ends up a _tiny_ bit blurry due to the browser resizing it. I think you can take a screenshot that is at the size the image will render at on this page (looks like 418x357) and make it super crisp.

brikis98 (approved):
These images are terrific and a big improvement 👍 

I think the PR can be merged as-is. However, I left some comments on possible improvements that can be done in follow-up issues / PRs.

> > Some of the rows could have a "refresh" icon (to indicate "in progress") and instead of "Open Pull Request," a link to an existing PR (e.g., "Pull Request Update the Landing Zone deployment guide to take into account the latest fixes/code in module-security #344" hyperlink).
> 
> Not sure if I follow, does the one below makes sense?
> 
> Screen Shot 2021-08-17 at 14 06 39

Ah, sorry I wasn't clear. I meant the icons in the "Status" column. Right now, they are all either a green check (which I assume means "up to date") or a red warning (which I assume means "needs to be updated"); I was suggesting adding a 3rd icon in that column which means "currently being updated", to indicate there's an open PR for that one already. 

> > Might be a good idea to have a mix of Gruntwork and non Gruntwork dependencies. E.g., Perhaps the list is something like:
> 
> Mixing different programming languages is confusing for me. I don't know how to present this information.

This is a NIT, and can be done in a follow-up PR, but probably the way to do it is:

1. Replace the column heading "Module" with "Dependency"
1. Add a column heading "Type" which can be set to "Terraform" or "JavaScript" or "Ruby" depending on the type of dependency. Perhaps we use the logos for those tools/languages rather than words to save space.
1. Now some of the rows can have Terraform modules, some can have JavaScript dependencies, some can have Ruby dependencies, etc.  

> Thoughts from a live conversation with Ina:
> 
> We will only release patchers for the important dependencies, e.g. React and Rails but not the `react-html-email` dependencies. For JS developers, they can use a [npm update](https://docs.npmjs.com/cli/v7/commands/npm-update) and then open a PR updating _all_ their dependencies, so having a button to update each dependency separately seems counter-intuitive.

I see this slightly differently:

1. You shouldn't have to run anything manually to update dependencies. Developers today can run `npm update` any time, but the reality is that they rarely do. The point of automatic dependency updates is to remove the need for devs to keep things up to date manually, whether the dependency is "small" or "large."
1. Most of the automatic updates are version number bumps; there is no patch necessary for those. This will work exactly as DependaBot or RenovateBot.
1. But when a dependency has released a new version that is backwards incompatible, then it's possible the maintainer has included a patch file in that release to make the upgrade easier. Thinking through this makes me realize two things:
    1. The table (in both the web UI and CLI output) should probably show if the new release has breaking changes or not.
    1. Similarly, it should show if a patch is available to handle those changes.
    1. Finally, it occurs to me that it won't always be the maintainer who creates a patch. Especially in the early days of Patcher, no one will know about it, including most maintainers of most projects. However, there will be a small group of devs who use Patcher and may want to create patches, _both for their own code and for code created by others_, to make updating easier. That group of devs will include Gruntwork employees—e.g., we may create patches for Terraform or K8S upgrades, even though we aren't the owners of Terraform or K8S—as well as Patcher customers, especially enterprise customers, who may want to create Patches for dependencies they don't own because they need a way to update those dependencies in hundreds of places in their code base. I don't think there is anything to change in the UI to reflect this now, but it's something to keep in mind for the future.

_data/patcher-landing-page-tabs-dependencies-list.yml
marinalimeira: Having this image for now would be better?

_data/patcher-landing-page-tabs-dependencies-list.yml
marinalimeira: Having this image for now would be better?

carbon

marinalimeira:
@brikis98 I made the text updates, will open a new PR with the improvements for the images. Can I get a ✅ ?

marinalimeira:
> I see this slightly differently:
> 
>  You shouldn't have to run anything manually to update dependencies. Developers today can run npm update any time, but the reality is that they rarely do. The point of automatic dependency updates is to remove the need for devs to keep things up to date manually, whether the dependency is "small" or "large."
>     Most of the automatic updates are version number bumps; there is no patch necessary for those. This will work exactly as DependaBot or RenovateBot, for all dependencies, "small" or "large."
>     But when a dependency has released a new version that is backwards incompatible, then it's possible the maintainer has included a patch file in that release to make the upgrade easier. Thinking through this makes me realize a few things:
>         The table (in both the web UI and CLI output) should probably show if the new release has breaking changes or not.
>         Similarly, it should show if a patch is available to handle those changes.
>         Finally, it occurs to me that it won't always be the maintainer who creates a patch. Especially in the early days of Patcher, no one will know about it, including most maintainers of most projects. However, there will be a small group of devs who use Patcher and may want to create patches, both for their own code and for code created by others, to make updating easier. That group of devs will include Gruntwork employees—e.g., we may create patches for Terraform or K8S upgrades, even though we aren't the owners of Terraform or K8S—as well as Patcher customers, especially enterprise customers, who may want to create Patches for dependencies they don't own because they need a way to update those dependencies in hundreds of places in their code base. I don't think there is anything to change in the UI to reflect this now, but it's something to keep in mind for the future.
>

Yes, this makes sense!! Thanks

marinalimeira:
>Ah, sorry I wasn't clear. I meant the icons in the "Status" column. Right now, they are all either a green check (which I assume means "up to date") or a red warning (which I assume means "needs to be updated"); I was suggesting adding a 3rd icon in that column which means "currently being updated", to indicate there's an open PR for that one already.

What do you think about this one?

Screen Shot 2021-08-18 at 17 21 36

brikis98:
> I am thinking now that with this "Open PR" we give the idea that its a manual process. Well, it's a 1-click process, but it still needs someone to go there and request to open a PR.

Ah, that's a good point. I think it's OK to merge the image you have above, but yea, perhaps the more clear UI for this in the future would be:

1. A table where there is one dependency per row, as you have it now.
1. For each dependency, you see the current version and latest available version. 
1. The status column has icons like ✅  ("up to date"), ⚠️ ("out of date"), or 🚨 ("known vulnerability"). Perhaps we even put that text next to each icon.
1. The last column should probably be labeled "actions" or similar. The contents could be ⚙️ ("change version") for dependencies up to date or 🔂  ("open pr #XXX") for out of date dependencies that are being automatically updated.

brikis98 (approved):

PR #585, opened 2021-08-13
Eks enterprise fixes

yorinasub17:
- https://gruntwork.atlassian.net/browse/WWW-125 (c29ff8a)
- https://gruntwork.atlassian.net/browse/WWW-126 (d7e8005)
- https://gruntwork.atlassian.net/browse/WWW-127 (0afce74)

### DIFF
--- a/_data/enterprise-landing-page-tabs-eks-generatecode.yml
+++ b/_data/enterprise-landing-page-tabs-eks-generatecode.yml
@@ -34,59 +34,62 @@
     <figure><figcaption>prod/us-west-2/services/eks_cluster/main.tf</figcaption>
       <pre>
       <code class="language-hcl">
-        # Example snippet from one of many Terraform files you get
+      # Example snippet from one of many Terraform files you get
 
-        module "eks_cluster" {
-          # Since CIS is enabled, source from the CIS compliant
-          # service catalog.
-          source  = "github.com/gruntwork-io/cis-service-catalog//eks-cluster"
-          version = "v0.51.0"
+      module "eks_cluster" {
+        # Since CIS is enabled, source from the CIS compliant
+        # service catalog.
+        source =
+          "github.com/gruntwork-io/cis-service-catalog//eks-cluster"
+        version = "v0.51.0"
 
-          cluster_name = "my-prod-cluster"
-          vpc_id       = dependency.vpc.outputs.vpc_id
+        cluster_name = "my-prod-cluster"
+        vpc_id       = dependency.vpc.outputs.vpc_id
 
-          managed_node_group_configurations = {
-            "prod-asg" = {
-              min_size   = 2
-              max_size   = 4
-              subnet_ids = dependency.vpc.outputs.private_app_subnet_ids
-              tags       = local.default_tags
-            }
+        managed_node_group_configurations = {
+          "prod-asg" = {
+            min_size   = 2
+            max_size   = 4
+            subnet_ids = dependency.vpc.outputs.private_app_subnet_ids
+            tags       = local.default_tags
           }
-
-          enable_pod_security_policies     = true
-          enable_network_security_policies = true
-          enable_eks_cluster_hardening     = true
         }
 
-        module "eks_core_services" {
-          source  = "github.com/gruntwork-io/cis-service-catalog//eks-core-services"
-          version = "v0.51.0"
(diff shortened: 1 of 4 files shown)


josh-padnick:
Thanks, Yori!

## WWW-125

Looks good!

## WWW-126

I'm not sure how usable this is:

image

I realize it would make the contact us form longer, but can you try stacking the checkboxes vertically to see if that improves readability?

## WWW-127

Looks good!

yorinasub17:
> I realize it would make the contact us form longer, but can you try stacking the checkboxes vertically to see if that improves readability?

Done: 
64061a7

josh-padnick (approved):
LGTM!

yorinasub17:
Thanks for review! Merging.

brikis98:
> ## WWW-126
> I'm not sure how usable this is:
> 
> image
> 
> I realize it would make the contact us form longer, but can you try stacking the checkboxes vertically to see if that improves readability?

Hm, I think the design I had in the [original ticket](https://gruntwork.atlassian.net/browse/WWW-126) was a bit different:

Screen Shot 2021-08-10 at 9 51 20 AM

Notes:

1. No "Which features are you interested in?" We just go straight into the options.
2. Each option fits on one row. I did that work in the browser, so I'm confident you can make those fit.

I think that had the best combination of readability and not making the form too long.

josh-padnick:
Jim, I asked Yori to make it vertical since I found it longer to scroll but easier to fill out. I actually still think that's the case and prefer what's currently there. That being said, why don't you make the final call here so we're not giving conflicting directives.

yorinasub17:
> Each option fits on one row. I did that work in the browser, so I'm confident you can make those fit.

Can you give me some pointers here on how you did that? No matter how I organize the grid, I can't get it as clean as you have it. Did you use bootstrap rows to do this? Or HTML tables in the row? Or raw padding and margins?

brikis98:
> Jim, I asked Yori to make it vertical since I found it longer to scroll but easier to fill out. I actually still think that's the case and prefer what's currently there. That being said, why don't you make the final call here so we're not giving conflicting directives.

I don't think it's a huge issue, but I find the long vertical page intimidating, and worry many customers will just bounce rather than filling that out. My guess is the horizontal layout is just as easy to understand, but less scary, in the sense that you can see the end of the page, rather than getting the impression that it's an endless form.

> > Each option fits on one row. I did that work in the browser, so I'm confident you can make those fit.
> 
> Can you give me some pointers here on how you did that? No matter how I organize the grid, I can't get it as clean as you have it. Did you use bootstrap rows to do this? Or HTML tables in the row? Or raw padding and margins?

That's my bad, I should've saved/shared the code when I was messing with it in dev tools. At this point, I don't entirely remember. I _think_ what I did was add `float: left` on the `div` that wraps each checkbox/label. I then set `width` on those `div` so that they were spaced out as equal sized columns: e.g., the first `div` in each row had `width: 120px`, the second div had `width: 160px`, etc. And I arranged the choices so that the shorter options went in narrower columns, longer options in wider ones, etc. It's ugly, but seemed to work. Oh, I also removed the top padding on the `label` for each row (e.g., Service Mesh, Self-service Dashboard) so it aligned better. I just repro'd it for a couple rows quickly in web tools and copy/pasted it out (the `style=` are the key changes):

```html
<div class="form-group">
  <div class="row">
    <div class="col-xs-12 col-md-4">
      <label style="padding-top: 0;">Service Mesh</label>
    </div>
    <div class="col-xs-12 col-md-8">
      <div style="float: left; width: 120px;">
        <input id="istio" name="interested-feature-service-mesh" type="checkbox" value="Istio">
        <label for="istio">Istio</label>
      </div>
      <div style="float: left; width: 160px;">
        <input id="consul-connect" name="interested-feature-service-mesh" type="checkbox" value="Consul Connect">
        <label for="consul-connect">Consul Connect</label>
      </div>
      <div style="float: left; width: 160px">
        <input id="aws-app-mesh" name="interested-feature-service-mesh" type="checkbox" value="AWS App Mesh">
        <label for="aws-app-mesh">AWS App Mesh</label>
      </div>
    </div>
  </div>
</div>
```

yorinasub17:
Thanks for the pointer. This is now done in https://github.com/gruntwork-io/gruntwork-io.github.io/pull/598

PR #586, opened 2021-08-13
Replace static argocd create image with gif

yorinasub17:
As suggested from https://gruntwork.atlassian.net/browse/WWW-124. I still have doubts about the gif, but after trying many different screenshot angles, the gif ended up being the most sensible to convey the self-service message.


josh-padnick:
Checked out the Argo image and it looks good! Approving.

josh-padnick (approved):

yorinasub17:
Thanks for review. Merging now!

PR #596, opened 2021-08-17
Small EKS page fixes

### LINKED ISSUE #592: Make the on-going value of the EKS subscription clearer
We got feedback from a user: https://twitter.com/aussielunix/status/1427589433323233286

> So my first reaction when I got to the page you linked was that it was not subscription, it was chose options, a swipe card, and a repo of code is handed over "100% mine"

After reading this, I realized that the "Amazon EKS for the enterprise. In about a day." headline makes it sound like this is a one-time action. This is problematic in two ways:

1. Cost: users may be surprised to learn it's an on-going subscription.
1. Value: users may not realize that we provide commercial maintenance & support, which is _critical_ to keeping all this stuff working. So we're positioning ourselves as a way to _launch_ an enterprise-grade EKS cluster, but not highlighting the value that this is also a way to _maintain_ an enterprise-grade EKS cluster.

### LINKED ISSUE #593: Less blurry images for EKS landing page
A few of the images on the EKS landing page are blurry:

1. Infrastructure Code
2. README.md
3. TFC
4. Stay up to date

For each one, we should do one of the following:

1. Replace with a less blurry image, if such an image is possible in such a tight space
2. Replace with a crisp `.gif`
3. Replace with a code snippet

brikis98:
1. Fix #593 / [WWW-124](https://gruntwork.atlassian.net/browse/WWW-124): replace several images that were a bit blurry with code snippets, `.gif`, or sharper images.
1. Fix #592: Add commercial maintenance & support to top-level list of features.
1. Make some code snippets a bit less wide to remove need for horizontal scrolling.

Preview URL: https://deploy-preview-596--keen-clarke-470db9.netlify.app/eks-for-enterprise


zackproser (dismissed):
LGTM

_data/enterprise-landing-page-tabs-eks-generatecode.yml
```diff
@@ -93,3 +87,68 @@
       </code>
       </pre>
       <figure>
+
+- id: app-deploy-code
+  title: sample-app/main.tf
```
yorinasub17, line 92: This is not what customers would get in the new world, as Argo and Flux do not support terraform. They would actually get something like this:

https://github.com/yorinasub17/gruntwork-sample-app-deploy/tree/main/aws-sample-app

Also, this conflicts with Step 5.

brikis98:
Doh, looks like my new `.gif` doesn't loop. Fixing...

yorinasub17 (commented):
LGTM except for the sample app code snippet.

NOTE: I am not a fan of the gifs but I am not really qualified to make that call. My concerns about gifs is mentioned in JIRA: https://gruntwork.atlassian.net/browse/WWW-124?focusedCommentId=13819

_data/enterprise-landing-page-tabs-eks-generatecode.yml
brikis98: Ohhh, good point, that completely overlaps! OK, will switch to VPC code snippet instead...

brikis98: Updated! https://github.com/gruntwork-io/gruntwork-io.github.io/pull/596/commits/d9c6cd3527bae1ea1af1ac16e7bdc7c7ac2037cb

yorinasub17 (approved):
LGTM

brikis98:
Thanks for the reviews! Merging and releasing now.

PR #611, opened 2021-08-20
Fix auto scaling setting for kops

brikis98:
Looks like it's available as a setting out-of-the-box: https://kops.sigs.k8s.io/addons/


yorinasub17 (approved):
LGTM!

brikis98:

PR #613, opened 2021-08-23
Update managed as code section in EKS comparison table

brikis98:
EKS isn't actually managed as code _out of the box_. That takes a bunch of extra work.


yorinasub17 (approved):
LGTM!

brikis98:

PR #622, opened 2021-08-23
Make featured headlines come before featured img

### LINKED ISSUE #616: Wide images in Patcher landing page break the flow a bit
The [Patcher home page](https://www.gruntwork.io/patcher) zig-zag now includes a few wide images: e.g., "Use the UI" and "You review and merge." I believe this was done so the images aren't shrunk by the browser and are more legible, and it's a lovely creative solution. However, the drawback to this is that it breaks the flow, making it hard to follow the page. The text now goes left-center-left-right-center, and what's a headline and what's not is thrown off. Users typically don't read a page, but skim headlines quickly (we've confirmed this from user testing too), so it's important the headlines are laid out in a consistent way that's easy to skim.

I can think of two options:

1. Go back to the original image sizes / zig-zag flow. Resize the images [as described here](https://gruntwork.atlassian.net/browse/WWW-124?focusedCommentId=13843) so they fit the available space, but aren't blurry.
1. Figure out a different way to do the "flow" for Patcher that works with large images, so that it's easy to skim down the page, spot headlines, etc.

Note also that these images show up on the [How it Works page](https://www.gruntwork.io/patcher/how-it-works), but at a different size (and as a result, blurry, as described in https://github.com/gruntwork-io/gruntwork-io.github.io/issues/619), so whatever solution we figure out for the Home Page should work for the How it Works page too.

zackproser:
Fixes #616


bwhaley (commented):
On the How it Works page, the third item in the Maintainer workflow (_Create a patch file_) is oriented backward, e.g. the code is on the left, the text is on the right, but it should be the opposite.

zackproser:
> On the How it Works page, the third item in the Maintainer workflow (_Create a patch file_) is oriented backward, e.g. the code is on the left, the text is on the right, but it should be the opposite.

Yeah, that's a long-standing issue with all of our how-it-works pages. I'll have to wrestle with the columns + offsets on that tomorrow.

brikis98 (commented):
Thanks for taking a crack at this Zack!

I'm really not trying to nitpick for the sake of nitpicking, but I have a gut feeling that the large images are making it harder to skim the page.

Here's an idea to try:

First, for the pull request image (the one below "You review and merge"), here's a smaller screenshot that I think will cleanly fit the zig-zag and still captures all the same content:

Screen Shot 2021-08-24 at 2 06 39 PM

_(I created it by shrinking my browser window down and using chrome dev tools to remove / shrink stuff that wasn't necessary)_

Now, the "Use the UI" and "Use the CLI" tabs are harder. Perhaps what we can do there is to create a thumbnail of each that fits the zig-zag UI (410x310) and overlay a "play" button on top of each (since something is overlayed on the thumbnail anyway, it doesn't have to be super crisp). When the user clicks the play button, we add a tiny bit of JS code to replace the image with the full-size, sharp, readable gif, which starts playing. That way, while skimming, you have the normal zig-zag flow, but when you want to see details, you can still do so, and you'll be paying full attention to it after clicking (rather than seeing a `.gif` part way through playing while scrolling).

WDYT?

zackproser:
> Thanks for taking a crack at this Zack!
> 
> I'm really not trying to nitpick for the sake of nitpicking, but I have a gut feeling that the large images are making it harder to skim the page.
> 
> Here's an idea to try:
> 
> First, for the pull request image (the one below "You review and merge"), here's a smaller screenshot that I think will cleanly fit the zig-zag and still captures all the same content:
> 
> Screen Shot 2021-08-24 at 2 06 39 PM
> 
> _(I created it by shrinking my browser window down and using chrome dev tools to remove / shrink stuff that wasn't necessary)_
> 
> Now, the "Use the UI" and "Use the CLI" tabs are harder. Perhaps what we can do there is to create a thumbnail of each that fits the zig-zag UI (410x310) and overlay a "play" button on top of each (since something is overlayed on the thumbnail anyway, it doesn't have to be super crisp). When the user clicks the play button, we add a tiny bit of JS code to replace the image with the full-size, sharp, readable gif, which starts playing. That way, while skimming, you have the normal zig-zag flow, but when you want to see details, you can still do so, and you'll be paying full attention to it after clicking (rather than seeing a `.gif` part way through playing while scrolling).
> 
> WDYT?

Makes sense - I will take a crack at implementing that.

zackproser:
Implemented in a5a16189

brikis98 (approved):
LGTM, thanks Zack!

brikis98:
Merging now.

PR #641, opened 2021-09-08
Replace usage of "fork" with "clone" in Terms of Service

josh-padnick:
Our docs recommend cloning our repos instead of forking them in the event you terminate your license so that you don't lose your forks once we revoke your access. This update makes our terms consistent with our docs.


bwhaley (approved):
LGTM!

josh-padnick:
Thanks, releasing now. You can share https://gruntwork.io/legal/feed with the customer in about 10 mins if you want to share the upadte.

PR #646, opened 2021-09-17
[APT-1373] Switch from Chargebee to Recurly for checkout.

ebeneliason:
In addition to the checkout service change, this tidies up a bit of CSS, fixes a bug that caused incorrect subscription totals from being shown on mobile, and replaces the $500 deposit with a "due now" total equal to the first month + fixed costs (i. e. RefArch), and supporting text to indicate the ongoing monthly fee.

We must wait to merge this until the production Recurly service is fully configured for use, the URL updated in this patch, and the new Recurly checkout zap turned on.

### DIFF
--- a/assets/js/checkout.js
+++ b/assets/js/checkout.js
@@ -135,26 +135,32 @@ $(function () {
   function _updateCheckout(newOptions) {
     checkoutOptions = Object.assign({}, checkoutOptions, newOptions);
 
-    var enable_pro_support = checkoutOptions.pro_support;
-    var setup_deployment = checkoutOptions.setup_deployment;
-    var setup_compliance = checkoutOptions.setup_compliance;
+    _updateUI();
+    _updatePrice();
+    _updateCheckoutLink();
+  }
+
+  function _updateUI() {
+    var support = checkoutOptions.pro_support;
+    var refarch = checkoutOptions.setup_deployment;
+    var compliance = checkoutOptions.setup_compliance;
 
     $('.grunty-sprite').attr('data-sprite', 0);
     $('#subscription_type').val(checkoutOptions.subscription_type);
 
     // updates addon switch
-    $('[data-switch]' + '[name="pro_support"]').prop('checked', enable_pro_support);
-    $('[data-switch]' + '[name="setup_deployment"]').prop('checked', setup_deployment);
-    $('[data-switch]' + '[name="setup_compliance"]').prop('checked', setup_compliance);
+    $('[data-switch]' + '[name="pro_support"]').prop('checked', support);
+    $('[data-switch]' + '[name="setup_deployment"]').prop('checked', refarch);
+    $('[data-switch]' + '[name="setup_compliance"]').prop('checked', compliance);
 
-    if (enable_pro_support) {
+    if (support) {
       $('.grunty-sprite').attr('data-sprite', 2);
       $('#subscription-addon-1').removeClass('check-list-disabled');
     } else {
       $('#subscription-addon-1').addClass('check-list-disabled');
     }
 
-    if (setup_deployment) {
+    if (refarch) {
       $('.grunty-sprite').attr('data-sprite', 1);
 
       $('#checkout-price-addon').show();
@@ -170,7 +176,7 @@ $(function () {
       $('#subscription-addon-2').addClass('check-list-disabled');
     }
 
-    if (setup_compliance) {
+    if (compliance) {
       $('#subscription-addon-3').removeClass('check-list-disabled');
 
       // Update Refarch texts to reflect CIS selection
@@ -184,22 +190,20 @@ $(function () {
       $('#addon-text-refarch').text('Reference Architecture');
     }
 
-    if (checkoutOptions.pro_support && checkoutOptions.setup_deployment) {
+    if (support && refarch) {
       $('.grunty-sprite').attr('data-sprite', 3);
     }
-
-    _calculatePrice();
   }
 
-  function _calculatePrice() {
-    var total, subtotal, subscriptionTotal;
+  function _updatePrice() {
+    var monthlyTotal, dueNowTotal;
 
     switch (checkoutOptions.subscription_type) {
       case 'aws':
-        total = subtotal = pricing.subscriptions.aws.price.value;
+        monthlyTotal = pricing.subscriptions.aws.price.value;
         break;
       case 'gcp':
-        total = subtotal = pricing.subscriptions.gcp.price.value;
+        monthlyTotal = pricing.subscriptions.gcp.price.value;
         break;
       default: // do nothing
     }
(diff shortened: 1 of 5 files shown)


brikis98:
I just took a quick look out of curiosity, and it looks like when you click checkout, you are taken to this page:

Screen Shot 2021-09-20 at 2 40 58 PM

A few thoughts/questions:

1. We used to have the checkout screen in a modal, which was nice, as it kept the context of gruntwork.io. Does Recurly not offer a modal experience?
1. The order summary shows a price of $0.00 for the "AWS Subscription (Monthly Payments)". That's quite confusing. Any way to hide that? Or to put the $795/mo there?
1. It's not clear the Ref Arch is a one-time fee. It gets added to the "order total" just like everything else. 
1. And for that matter, it's not clear that "Users" at $795, "CIS AWS Foundations Benchmark Compliance" at $1,245.00 and "Pro Support" at $1,950.00 are per month fees. The only thing that mentions "month" is "Monthly Payments," which as I wrote above, is at $0.00. So perhaps users will assume everything here is a one-time fee.
1. Being able to enter how many users you want is pretty cool, but there are a few oddities:
    1. When you go over 30, it goes to $999,999/mo. I assume this means "contact us for enterprise"? If so, is there any way to make that clearer? I suspect most users will get it, but then they'd still have to figure out how to contact us, and a few users might just get confused and think we don't support 30 users or our checkout form is broken.
    1. The ability to enter less than 20 users may tempt some users into trying that to see if they'll save money. I guess the price doesn't go below $795/mo even if you enter 0 users, but again, it's not super obvious what it means to enter 0 here. 
1. On step 2, we have the "I accept the Privacy Policy and Terms of Service" checkbox, but not the "I understand I cannot resell Gruntwork services without a separate agreement" checkbox.

ebeneliason:
For some context, the Recurly hosted checkout page shown here is a temporary solution for us. We intend to eventually make use of Recurly.js to create a fully branded and customized checkout experience where we have full control, but dropping in the their hosted version unblocks a migration to Recurly, which has other benefits across the board. The Finance team, in particular, is keen to make this change at the beginning of the next quarter, on Oct 1, for bookkeeping reasons.

> The order summary shows a price of $0.00 for the "AWS Subscription (Monthly Payments)". That's quite confusing. Any way to hide that? Or to put the $795/mo there?

Josh will need to answer this. The fee shown there does change with the number of users using a stair-step pricing schema. I do wonder whether we could bake the base subscription cost into the line item above and make the first stair-step tier "free" in order to make this more clear.

> It's not clear the Ref Arch is a one-time fee. It gets added to the "order total" just like everything else.

Yes, indeed. This is a major limitation of the hosted checkout page. In fact, their hosted checkout platform doesn't actually support one-time fees _at all_. We're crudely circumventing this by virtue of the fact that their platform actually does a really bad job of communicating that this is a monthly recurring fee at all, anywhere on the page. We're essentially adding the RefArch as a recurring subscription, and the finance team will simply remove it as a recurring fee after the customer signs up, before they get their next bill, as a stop gap until we build our own checkout page.

This is also the reason that we've made changes to the communication on our checkout page, before reaching Recurly checkout. We want to make it very clear that there's an up-front cost, and a lower recurring subscription cost. The cost shown as a total on the Recurly checkout page matches the up front cost we display before they get there.

> And for that matter, it's not clear that "Users" at $795, "CIS AWS Foundations Benchmark Compliance" at $1,245.00 and "Pro Support" at $1,950.00 are per month fees. The only thing that mentions "month" is "Monthly Payments," which as I wrote above, is at $0.00. So perhaps users will assume everything here is a one-time fee.

Agreed. Again, we have no control over this, currently. Their platform is surprisingly ambiguous about these details.

> We used to have the checkout screen in a modal, which was nice, as it kept the context of gruntwork.io. Does Recurly not offer a modal experience? Without it, all the context from our own page (e.g., about what's a monthly payment or not, our free trial, etc) is no longer visible.

No, I don't believe they support that. Here are the docs: https://docs.recurly.com/docs/hosted-payment-pages

> Being able to enter how many users you want is pretty cool, but there are a few oddities:

> When you go over 30, it goes to $999,999/mo. I assume this means "contact us for enterprise"? If so, is there any way to make that clearer? I suspect most users will get it, but then they'd still have to figure out how to contact us, and a few users might just get confused and think we don't support 30 users or our checkout form is broken.

> The ability to enter less than 20 users may tempt some users into trying that to see if they'll save money. I guess the price doesn't go below $795/mo even if you enter 0 users, but again, it's not super obvious what it means to enter 0 here.

Agreed. I think this is essentially the best we could do given the constraints of the platform (until we switch to Recurly.js — this is the big theme here), but I'll let @josh-padnick comment on that as well.

> On step 2, we have the "I accept the Privacy Policy and Terms of Service" checkbox, but not the "I understand I cannot resell Gruntwork services without a separate agreement" checkbox.

Correct. I discussed this with Josh and he felt that may no longer be necessary, and it's also covered by the ToS itself. Maybe that's still a concession, given that there is no way to add such a checkbox to the Recurly checkout page. We could add one to our website, pre-checkout, but it wouldn't be "enforceable" in the sense that there's no way to pass any record of the user's checking of that box through to our contract on the backend, so it would just be "for show."

assets/js/checkout.js
```diff
@@ -184,109 +190,87 @@ $(function () {
...
 
+    // CIS Compliance is only on AWS for now
+    if (checkoutOptions.setup_compliance) {
+      monthlyTotal += pricing.subscriptions.aws.cis_compliance_price.value;
+    }
+
+    dueNowTotal = monthlyTotal;
+
     // Only AWS supports the Ref Arch
     if (checkoutOptions.setup_deployment) {
```
eak12913, line 231: It's a little confusing because `checkoutOptions` still retains the older naming: `setup_deployment` vs `refarch` as you've named it here. Your naming is more up to date. Probably doesn't make sense to pay down the tech debt in this PR though..

assets/js/checkout.js
```diff
@@ -184,109 +190,87 @@ $(function () {
...
-      total += subtotal = pricing.subscriptions.aws.cis_compliance_price.value;
+    $('#due-now').text(dueNowTotal.toLocaleString());
+    $('.monthly-total').text(monthlyTotal.toLocaleString());
+  }
+
+  // Update the Recurly checkout link URL
+  function _updateCheckoutLink() {
+    const type = checkoutOptions.subscription_type;
+    const support = checkoutOptions.pro_support;
+    const setup = checkoutOptions.setup_deployment;
```
eak12913, line 247: maybe call this `refarch` to be consistent with above naming?

assets/js/checkout.js
```diff
@@ -184,109 +190,87 @@ $(function () {
...
+      add_on_quantity: Object.values(addOns).toString(),
+    };
 
-    _deferCheckout(checkoutOptions.subscription_type,
-      checkoutOptions.pro_support,
-      checkoutOptions.setup_deployment,
-      checkoutOptions.setup_compliance);
-  }
+    for (const key in params) {
+      href += "&" + key + "=" + params[key]
```
eak12913, line 270: Does this not end up generating a url like: `"......-monthly?&users=20"`? Note the `&` after the `?`

ebeneliason: It does. It's avoidable, but makes the code a little uglier. (I actually handled it at first, then took it out to simplify the logic.) It doesn't impact the behavior to begin with the separator — does it explicitly go against the spec or fail to work in some contexts? I'm not sure.

eak12913: Hmm it does produce such a URL: 
image

Usually you just see `?firstParam=value&nextParam=anothervalue.....` Not sure if that's a legit problem or just a technicality

ebeneliason: Yeah, exactly. It works fine in my experience, but I agree it's slightly unconventional. Would this satisfy you more? It's no big deal to add it, so maybe we should just conform to convention.
```
    sep = "";
    for (const key in params) {
      href += sep + key + "=" + params[key];
      sep = "?";
    }
```

ebeneliason: Pushed this change!

assets/js/checkout.js
ebeneliason: Yeah, agreed. We can clean up the tech debt when we switch over to a custom Recurly.js page. In the meantime, the shorthand local variables offer a decent and consistent way to map them.

ebeneliason:
@brikis98 We've made a few adjustments to the page to improve clarity within the bounds of what's possible…
* The base subscription now displays as $795
* The users line item no longer appears at all (We'll continue defaulting to 20 users, non-configurable at checkout for now)
* The Reference Architecture line item now has a "(One-Time)" qualifier to distinguish from the other monthly fees

We think this addresses the most urgent concerns sufficiently to move forward with this hosted payment approach, again with the caveat that it's a temporary solution and we hope to ship a customized checkout process within the coming month. Do these changes mitigate your concerns?

josh-padnick:
@brikis98 FYI, we plan to launch this in the next day or so. If you have any further comments, please share them with ASAP!

brikis98:
> For some context, the Recurly hosted checkout page shown here is a temporary solution for us. We intend to eventually make use of Recurly.js to create a fully branded and customized checkout experience where we have full control, but dropping in the their hosted version unblocks a migration to Recurly, which has other benefits across the board. The Finance team, in particular, is keen to make this change at the beginning of the next quarter, on Oct 1, for bookkeeping reasons.

I'm usually in favor of MVPs and incremental improvement, but some of the changes here are very, very confusing, and I'm worried will result in a drop in sales or really pissed off customers. So I think we need to be careful here. Let me catch up on all your other comments and follow-up replies too.

> 
> > The order summary shows a price of $0.00 for the "AWS Subscription (Monthly Payments)". That's quite confusing. Any way to hide that? Or to put the $795/mo there?
> 
> Josh will need to answer this. The fee shown there does change with the number of users using a stair-step pricing schema. I do wonder whether we could bake the base subscription cost into the line item above and make the first stair-step tier "free" in order to make this more clear.
> 
> > It's not clear the Ref Arch is a one-time fee. It gets added to the "order total" just like everything else.
> 
> Yes, indeed. This is a major limitation of the hosted checkout page. In fact, their hosted checkout platform doesn't actually support one-time fees _at all_. We're crudely circumventing this by virtue of the fact that their platform actually does a really bad job of communicating that this is a monthly recurring fee at all, anywhere on the page. We're essentially adding the RefArch as a recurring subscription, and the finance team will simply remove it as a recurring fee after the customer signs up, before they get their next bill, as a stop gap until we build our own checkout page.
> 
> This is also the reason that we've made changes to the communication on our checkout page, before reaching Recurly checkout. We want to make it very clear that there's an up-front cost, and a lower recurring subscription cost. The cost shown as a total on the Recurly checkout page matches the up front cost we display before they get there.
> 
> > And for that matter, it's not clear that "Users" at $795, "CIS AWS Foundations Benchmark Compliance" at $1,245.00 and "Pro Support" at $1,950.00 are per month fees. The only thing that mentions "month" is "Monthly Payments," which as I wrote above, is at $0.00. So perhaps users will assume everything here is a one-time fee.
> 
> Agreed. Again, we have no control over this, currently. Their platform is surprisingly ambiguous about these details.

Oof... When it comes to pricing, we need to be _crystal clear_ on what's recurring and what's one-time. Customers are _already_ surprised by our recurring fees and annual contracts. It could be very harmful to have a misleading checkout form making these problems worse.

OK, I'm going to hit "Comment" and then go on to respond to subsequent replies.

brikis98:
> @brikis98 We've made a few adjustments to the page to improve clarity within the bounds of what's possible…
> 
> * The base subscription now displays as $795
> * The users line item no longer appears at all (We'll continue defaulting to 20 users, non-configurable at checkout for now)
> * The Reference Architecture line item now has a "(One-Time)" qualifier to distinguish from the other monthly fees
> 
> We think this addresses the most urgent concerns sufficiently to move forward with this hosted payment approach, again with the caveat that it's a temporary solution and we hope to ship a customized checkout process within the coming month. Do these changes mitigate your concerns?

OK, updated version looks much clearer 👍 

Screen Shot 2021-09-28 at 12 42 57 PM

A few thoughts:

1. The "(Monthly)" and "(One-Time)" help quite a bit with clarity. Do we have space for a little more to make it even clearer? In particular, should it say "(monthly fee, annual contract)" for the recurring charges to make it extra clear it's an annual contract? Or even "(monthly fee, annual contract, 30 day money back guarantee)"? That last one is prob too damn long 😁 
1. Our website today only charges a $500 deposit. Now, if you select all the options, we bill you as much as $8,940.00 up front. I know that, in net, it's all the same, but I wonder if this is psychologically painful and might hurt sales? 

Overall though, it's much better, and addresses most of my concerns.

josh-padnick:
> Our website today only charges a $500 deposit. Now, if you select all the options, we bill you as much as $8,940.00 up front. I know that, in net, it's all the same, but I wonder if this is psychologically painful and might hurt sales?

Funny, I had the exact same concern and Eben brought up the fact that there's no avoiding showing the customer the price they just selected, so I guess we'll see what happens.

That being said, the good news is that this is still a stopgap solution. In the next sprint, we plan to start working on a fully branded checkout. That means we should launch it in ~6 weeks. So our blast radius is limited to just a few weeks of the hosted checkout before we can create a fully customized experience.

> In particular, should it say "(monthly fee, annual contract)" for the recurring charges to make it extra clear it's an annual contract?

@ebeneliason We should carefully review the checkout to see if this part is obvious. I guess nothing has really changed in our UX since the ChargeBee modal did nothing to clarify the annual contract part, but still, we don't want anyone to not realize they're committing to an annual contract.

ebeneliason:
> I guess nothing has really changed in our UX since the ChargeBee modal did nothing to clarify the annual contract part, but still, we don't want anyone to not realize they're committing to an annual contract.

Yeah, we'll have opportunity to do better here with our custom version. I did just check out the ChargeBee flow, and you're right that there's no mention of this at all (nor, for that matter, about any one-time _or_ monthly recurring cost). In that regard, this temporary hosted version is much more clear than the current flow.

That comes at the price, as you both point out, of being really transparent about the total cost. While in my mind there's no choice about being clear about that cost, we do have a choice with regard to what we ask them to pay up front. If it improved sales, we could even consider no up front cost, and avoid having to refund anything if they cancel within 30 days. Even if that were true, we'd have to make it very clear what the total cost of the subscription is.

josh-padnick:
@ebeneliason Agreed. Let's focus on launching the Recurly checkout and see how it performs!

josh-padnick (dismissed):

josh-padnick (dismissed):

josh-padnick (approved):

PR #651, opened 2021-10-15
Add Software Engineer (Contractor)

marinalimeira:
➕  Remove SRE.
➕ This job description will be the same as Workable and UpWork.

Preview: https://deploy-preview-651--keen-clarke-470db9.netlify.app/careers/#software-engineering-contract

### DIFF
--- a/pages/careers/_open-positions.html
+++ b/pages/careers/_open-positions.html
@@ -25,11 +25,6 @@
     <h4>What You'll Work On</h4>
     <ul>
       <li>
-        <strong>Document team knowledge:</strong> Capture siloed knowledge and
-        create knowledgebase articles, runbooks, and other documentation for the
-        internal team as well as Gruntwork customers.
-      </li>
-      <li>
         <strong
           ><a href="/infrastructure-as-code-library/"
             >Infrastructure as Code Library</a


brikis98:
> ➕ Remove SRE

Thanks!

> * Is ok having this position as "Senior"? My conclusion from reading the requirements is that we do not want someone mid-level.

Yep, I think that works.

> * Can the same job description go to Workable and UpWork?

Yep, I think so.

brikis98:
> > * Is ok having this position as "Senior"? My conclusion from reading the requirements is that we do not want someone mid-level.
> 
> Yep, I think that works.

Ah, on second thought, I just realized how we show the full-time software engineer job:

Screen Shot 2021-10-19 at 11 03 13 AM

So we have "Software Engineering" as the top-level grouping, and then the various titles (Software Engineer, Senior Software Engineer, etc) below it. 

We should probably follow this pattern for the contractor job, making "Software Engineering (contract)" as the top-level grouping and having Senior Software Engineer and Principal Software Engineer as the two titles we have open positions for.

marinalimeira:
> Please make sure to check out the description in the contractor RFC, as a bunch are missing here: e.g., the mention of a path to full-time work, the detailed list of responsibilities, and all the required / optional qualifications.

Thanks! I will check on that again and pay attention to these topics.

PR #660, opened 2021-11-11
Update enterprise users tooltip

josh-padnick:
Note that I accidentally committed ac9e14d1ae21e4457c9d6422aa546a96c259a7d4 directly to `master` so this PR is a diff on that commit.


ebeneliason (approved):
LGTM.

PR #666, opened 2021-12-01
Add Patcher images

ina-stoyanova:
Adding patcher images here so we can use them for Patcher. Otherwise we have to host these images somewhere else (Imgur, S3, or Github even)


marinalimeira (approved):
LGTM!

marinalimeira (approved):

PR #688, opened 2022-02-24
Remove vestigial $500 from terms of service

josh-padnick:


josh-padnick:
This PR is ready to merge, so a stamp of approval would be much appreciated!

PR #693, opened 2022-03-12
Update circleci image from deprecated circleci class to cimg

yorinasub17:
NOTE: Is there a way to test that this works correctly without merging to `master`? As far as I can tell, I don't think there is a way to do this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ orbs:
 references:
   base_container: &base_container
     docker:
-      - image: circleci/ruby:2.7.1-buster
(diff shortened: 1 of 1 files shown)


eak12913:
> NOTE: Is there a way to test that this works correctly without merging to `master`? As far as I can tell, I don't think there is a way to do this.

I think you'd have to change the CircleCI config _in this branch_ to make it run on _this_ branch (instead of `master`) and then remember to change it back if things looked good. Since this is a very simple change, I will just merge this and if there's an issue - I will revert.

eak12913 (approved):

PR #701, opened 2022-04-03
Update job postings

josh-padnick:
- Fix time zone issues
- Change from hiring "Director of Engineering" / "VP of Engineering" to Engineering manager. Based on the applicants we received, I think we're looking for someone more hands on.


josh-padnick:
Merging for now unilaterally; feedback always welcome.

PR #713, opened 2022-05-06
Update terms of service to remove security vulnerability services

josh-padnick:
I propose our email announcement to customers read as follows:

> Effective as of May 5, 2022, we are making changes to our Terms of Service. No action is needed from you today. We updated the following:
> - **Discontinued Security Vulnerability Notifications.** Historically, we emailed customers about critical security vulnerabilities, often just 2 or 3 per year. But since we began offering this service, many third parties have emerged that provide a better version of this functionality, in many cases at no cost. As one example, see [Vulmon Alerts](https://alerts.vulmon.com/). In addition, customers have told us they do not receive much value from these notifications, so we are discontinuing this courtesy service.
>
> [View Terms](https://gruntwork.io/legal/terms)
> Thank you for being a Gruntwork user!

@brikis98 Let me know what you think of the above.

### DIFF
--- a/_data/legal.yml
+++ b/_data/legal.yml
@@ -182,3 +182,10 @@
   category: vulnerability-disclosure-policy
   link: https://github.com/gruntwork-io/gruntwork-io.github.io/commit/2d765b5
   date: 2022-05-02
+
+- guid: 10022
+  title:  Removed our security alerts service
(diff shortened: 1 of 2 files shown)


brikis98 (approved):
LGTM!

PR #717, opened 2022-05-10
Update job titles

josh-padnick:
The EM role is more likely a Director of Engineering, so I've re-titled it accordingly.


josh-padnick:
Trivial change and I need it timely so merging!

PR #720, opened 2022-06-07
Remove bug bounty

ebeneliason:
We've removed the bug bounty from our Vulnerability Disclosure Policy due to the greatly decreased signal-to-noise ratio in the reports we received.


josh-padnick (approved):
LGTM!

PR #726, opened 2022-06-14
Remove consul, nomad, and vault modules and mentions

ebeneliason:
These modules are no longer supported.


_data/library.yml
```diff
@@ -648,71 +648,6 @@
...
-      blurb: install Vault on a Linux server
-      description: Install Vault and its dependencies on a Linux server.
-    - name: run-vault
-      blurb: configure and run Vault
-      description: Run Vault and automatically connect to a Consul cluster.
-    - name: private-tls-cert
-      blurb: generate a self-signed TLS cert
-      description: Generate self-signed TLS certificates for use with Vault.
-
 - name: "Couchbase"
```
yorinasub17, line 651: Couchbase can also be removed because our contract expired.

yorinasub17 (approved):
LGTM! Also flagged one more repo that can be removed.

_data/library.yml
ebeneliason: Thanks, pushed a new commit to remove Couchbase as well.

PR #728, opened 2022-07-07
[APT-1875] Switch Stripe checkout to product ids

eak12913:
This PR updates our Stripe checkout to use product ids rather than specific price ids. Aperture will then look up the product Ids in Stripe and will display and charge the customer the default prices associated with each product the customer intends to purchase.


ebeneliason (approved):
Short and sweet.

PR #729, opened 2022-07-07
[DO NOT MERGE][APT-1876] Switch all prod codes to live values

eak12913:
This PR updates all the product ids from "test" mode values to "live" values.


ebeneliason (commented):
I feel like I should sanity check all these values in Stripe just to be double sure of them before approving. In the meantime, a couple of comments to update.

ebeneliason:
We've made this update directly on the base branch due the conflicts encountered here.

PR #731, opened 2022-07-08
[APT-1875] Extract product data into yaml

eak12913:
This PR extracts the product data into a shared YAML file that both javascript and HTML files can use.


ebeneliason (approved):
This looks great!

Note that we'll need to update the PR which switches these to live values.

PR #745, opened 2022-09-13
pro-support.html: clarify pro support

iangrunt:


zackproser (approved):
LGTM - thanks for making this change!

PR #749, opened 2022-09-27
Fix social icon alignment

ebeneliason:


brikis98 (approved):
Oh, thank you! I kept seeing this, kept meaning to file a bug, but then immediately forgot about it 😁

PR #753, opened 2022-10-13
Remove Jenkins from Ref Arch pages

pete0emerson:
I am leaving Jenkins in place in areas that are not Ref Arch specific.


zackproser (approved):
LGTM

PR #765, opened 2022-12-12
Restore 'in about a day' messaging

bethadele:
We've been delivering the Ref Arch in about a day on average, and we're not certain that the "in a few days" messaging isn't eroding our marketing message, so we're restoring the "in about a day" wording throughout the site.


josh-padnick (approved):
LGTM!

PR #769, opened 2023-01-11
Updated the CircleCI job to use contexts 

eak12913:
This will make it easier to rotate credentials in the future.


pete0emerson (approved):
LGTM

PR #774, opened 2023-03-17
contact: update Contact Us form fields

iangrunt:


pages/contact/_hero.html
```diff
@@ -1,9 +1,8 @@
 <div class="container">
   <div class="row">
     <div class="col-xs-12 text-center">
-      <h1>Ask a Grunt</h1>
-      <p style="margin-bottom:4px" class="lead">Speak to a real human!</p>
-      <p>Use the form below or send an email to <a href="mailto:sales@gruntwork.io">sales@gruntwork.io</a>.</p>
+      <h1>Book a Gruntwork Demo!</h1>
+      <p style="margin-bottom:4px" class="lead">Use this form to book a live demo with our DevOps engineers.</p>
```
josh-padnick, line 5: You know, I think we should actually delete the `<p>` tag here. It looks even more clean without it.

image

pages/contact/_form.html
```diff
@@ -61,12 +41,12 @@
     <button
       type="submit"
       id="submit-button"
-      class="btn btn-primary"
+      class="col-xs-12 btn btn-primary"
```
josh-padnick, line 44: The long button looks a little weird. Perhaps the following would better?

image


If you like that idea, here's HTML to render it, but note that I have _not_ tested this code in any way.

```html
<div class="form-group">
    <div class="row">
      <div class="col-xs-12 col-md-4">
        <label for="contact-message"></label>
      </div>
      <div class="col-xs-12 col-md-8">
        <div class="text-center" style="text-align:right;display: block;">
    <br>
    <button type="submit" id="submit-button" class="col-xs-8 btn btn-primary" ga-on="click" ga-event-category="contact" ga-event-action="submit" style="
">
      Book a Demo
    </button>
  </div>
      </div>
    </div>
  </div>
```

josh-padnick (approved):
I'm so glad you took the lead on this. This looks way better. I left a couple comments, but LGTM!

PR #775, opened 2023-03-18
contact: Update Contact Us form

iangrunt:
Incorporate @josh-padnick's feedback from #774

### DIFF
--- a/pages/contact/_form.html
+++ b/pages/contact/_form.html
@@ -36,7 +36,7 @@
       </div>
     </div>
   </div>
-  <div class="text-center">
+  <div class="text-right">
     <br />
     <button
       type="submit"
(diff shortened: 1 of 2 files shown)


josh-padnick (approved):
LGTM (provided it works in mobile, which I didn't test).

Separately, I wonder if we should change "Contact Sales" in the nav to "Book a Demo."

Also, I wonder if we should just embed a meeting scheduler like https://www.hubspot.com/products/sales/schedule-meeting or chilipiper.com directly on our page? Chilipiper can do lead qualification for us, and then people can just book meetings. 

But let's consider that for another time. For now, this looks great. Feel free to make whatever remaining changes you think make sense and ship.

PR #776, opened 2023-03-20
Collapse checkout to a single step/page

brikis98:
Direct link to preview: https://deploy-preview-776--keen-clarke-470db9.netlify.app/pricing

Screenshot:

checkout

1. Instead of a two step checkout process, it's now just one step.
2. Clearly position Ref Arch, Pro Support, etc as add-ons.
3. Show starting enterprise pricing.

See [this thread](https://gruntwork-io.slack.com/archives/C048B71K9GV/p1678473327249339?thread_ts=1675791616.097539&cid=C048B71K9GV) for context.


brikis98:
@eak12913 @oredavids How do I test the "checkout" button? It just redirects me to Aperture for some reason?

pages/pricing/index.html
```diff
@@ -3,9 +3,13 @@
 title: Become a Gruntwork Subscriber
 excerpt: Choose the plan just right for your team.
 permalink: /pricing/
-redirect_from: /pricing-beta/
+redirect_from:
+  - /pricing-beta/
+  - /checkout/
```
brikis98, line 8: The old checkout page now redirects to this one-page version.

brikis98:
> First - if you sign in to https://aperture.dogfood-stage.com/ then you will be able to see the checkout page from your preview environment when you hit "Checkout". This should let you test out to make sure that you've properly implemented the transition between the https://gruntwork.io pricing page and then make sure that the right products make their way over to the checkout page in Aperture.

Oh, why do you need to be logged in for that? On the prod site you of course don't need login, as people going through checkout aren't customers yet. Is this for security reasons?

I added it to the README: https://github.com/gruntwork-io/gruntwork-io.github.io/pull/776/commits/9ca2b6103c280ecd3c05392046e58b041e59a537

> 
> Second - there's a problem with your logic right now. It works well for selecting additional products and then checking out. Unfortunately, if I tried to edit the stuff I selected then things break. To repro:
> 
> 1. Go to your preview env's pricing page and get the base subscription with pro support and a ref arch.
> 2. Hit Checkout - observe that you're now in aperture - and your cart contains exactly the two product addons.
> 3. Hit "edit" from Aperture's checkout page. It will send you back - now - none of your products are selected.
> 
> This is a regression from what we have deployed. Further - it's even worse if you don't click the "edit" link but just use the "back" button on your browser because in the case of using your browser's back button - you'll still see the old things "checked" (as you expect) but the underlying js doesn't really have that value. So if you had pro support and a ref arch and then hit "back" and select CIS and a ref arch - when you hit "Checkout" you'll end up on Aperture's checkout page and will only see just the _last_ think you checked (in my case it was only CIS)

I think this is now fixed by: https://github.com/gruntwork-io/gruntwork-io.github.io/pull/776/commits/16f033a8bc63b51a90f85b68b9bb3375b41e76fc

eak12913:
>Oh, why do you need to be logged in for that? 

We purposely did this to make it a little bit more robust to spam and benevolent "researchers" spamming us with various attempts. This way - only Grunts can really test the checkout flow.

eak12913 (dismissed):
LGTM!

brikis98:
> I've verified that things look good from the usability side of things and issue I reported above is now fixed.
> 
> The only other bit of feedback I've got is that we should consider being consistent with naming of our compliance product between the pricing page and what ends up in your cart. For example - in the pricing page, I can choose the "Compliance" product - but what ends up in the cart is "CIS Package".

Oh, good point! I've updated the titles to more closely match, plus cleaned up a few descriptions: https://github.com/gruntwork-io/gruntwork-io.github.io/pull/776/commits/74273418b0444c06ce1ffaa95238b47bb621516f

Could I get one more ship it plz?

pete0emerson (approved):
LGTM :shipit:

brikis98:
Thank you for the review! Merging now.

PR #780, opened 2023-03-28
Add product manager role

josh-padnick:
We're hiring a new product manager!

### DIFF
--- a/pages/careers/_open-positions.html
+++ b/pages/careers/_open-positions.html
@@ -21,8 +21,13 @@
     <h2>Software Engineering</h2>
 
     <h3 id="principal-software-engineer-iac" style="margin-top: 5px; margin-bottom: 5px">Principal Software Engineer - IaC Library &amp; Platform</h3>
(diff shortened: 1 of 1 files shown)


ebeneliason:
* Some of the deliverables described as project management practices don't fall entirely under PM purview, but I think that's probably okay since that will be collaborative among product/design/eng.
* Should we elevate the need for customer communication to a top level bullet? I assume this hire will play a critical role in the new customer success workflows.
* There's no mention of patcher or the EU team, and coordinating product and roadmaps across time zones will be required for success.
* There aren't a lot of hard requirements listed. Is that by design? Would listing some hard requirements for experience or capabilities make the screening process easier?
* There are some related items, but some key words like "roadmap" and "prioritization" aren't in the post. Not sure if that matters, just pointing it out.
* Nit: The time zone badge doesn't match the format of the other listings
* Nit: The headers in the new listing are mixed title/sentence case; other listings use sentence case consistently
* Nit: Use of prime instead of apostrophe in e.g "What you’ll work on"

ebeneliason (dismissed):
I'll go ahead and approve this because nothing is a dealbreaker, but I left some comments for your consideration.

ebeneliason (approved):

PR #781, opened 2023-03-29
thanks: update content and styling

iangrunt:
Update the page customers hit after booking a demo clearly next steps and better content.

### DIFF
--- a/pages/thanks/index.html
+++ b/pages/thanks/index.html
@@ -1,24 +1,73 @@
 ---
 layout: default
-title: Thank you.
-excerpt: Thank you for contacting us. A grunt will be in touch soon!
+title: "Gruntwork: Thanks!"
 permalink: /thanks/
+slug: thanks
 ---
 
+<style>
+  .thanks-text h1 {
+    color: #2C3E50;
+    margin-bottom: 15px;
+    text-align: center;
+  }
+
+  .thanks-text h3 {
+    color: #34495E;
+    margin-bottom: 15px;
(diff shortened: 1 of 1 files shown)


brikis98 (commented):
Looks great overall!

Only issue is the header looks wrong. It's all faded out... Some background is missing or something?
Screen Shot 2023-03-30 at 4 09 38 PM

eak12913 (commented):
I'm with Jim on this one - I understand we want to show them the additional resources - but it still looks _odd_. 

The action that's happened is they requested to speak to someone and are now being thanked for it. I don't think that fading out the text will be more likely to make the user watch the video. 

When I see "...in the meantime...we've gathered some useful resources..." type text - I give maybe a 2 seconds more of attention just to see if the answer to what I specifically need is in the list. There's about .01% chance that the answer is there..so I'm likely to just bounce.

brikis98:
> I'm with Jim on this one - I understand we want to show them the additional resources - but it still looks _odd_.
> 
> The action that's happened is they requested to speak to someone and are now being thanked for it. I don't think that fading out the text will be more likely to make the user watch the video.
> 
> When I see "...in the meantime...we've gathered some useful resources..." type text - I give maybe a 2 seconds more of attention just to see if the answer to what I specifically need is in the list. There's about .01% chance that the answer is there..so I'm likely to just bounce.

+1... The faded thing attracts my attention more and makes me wonder what's happening. Is there some modal or something?

PR #782, opened 2023-03-29
contact: update form and content

iangrunt:
Update the contact form based on feedback.


brikis98 (commented):
Content looks great, but the layout needs some help! Maybe @ebeneliason can lend a hand?

PR #783, opened 2023-03-30
Patcher

lubiarz-oliwia:
- patcher landing page (march),
- navbar updates,
- early access form (_a new form bucket link for the early access form is needed_),
- great news page,
- home updates,
- select plan page updates, 
- navbar updates, 
- achieve compliance page update,
- reference architecture page updates, 
- gruntwork landing zone page updates, 
- how it works page updates

### DIFF
--- a/assets/css/pages/patcher.scss
+++ b/assets/css/pages/patcher.scss
@@ -1,5 +1,4 @@
 .patcher {
-
   // Hero styles for the patcher page
   .section-hero {
     padding-top: 90px;
@@ -7,7 +6,6 @@
     @media (max-width: $screen-md-max) {
       padding-top: 36px;
       padding-bottom: 0px;
-
       .col-lg-7 {
         min-height: 225px;
       }
@@ -42,7 +40,6 @@
     h2 {
       margin-bottom: 0;
       margin-top: 0;
-
       &:last-child {
         a {
           display: none;
@@ -65,7 +62,6 @@
       tr {
         &:first-child {
           td {
-
             &:nth-child(2),
             &:nth-child(3) {
               border-top: 1px solid $input-border;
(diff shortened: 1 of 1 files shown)


josh-padnick:
I am strongly opposed to including future roadmap commitments. This gains us almost nothing ("Oh cool, they'll eventually do more things") but publicly commits us to a roadmap, or sets us up to fail to deliver what we say we will. Strong recommendation to remove this section entirely.

image

That being said, we _should_ encourage people interested in joining a paid beta to fill out some kind of conversion form so we can reach out to them about developing Patcher with them.

josh-padnick:
This page needs a lot more refinement:

image

Feedback:

- Visually, can we put the form on the right and content on the left? Right now, there's a lot to read before you get to the form.
- The form should be more detailed. We're not looking to drive high-volume leads at this stage, but to understand our customer better. Here are the fields I'd like to see on the form:
   - Name
   - Company email
   - Are you a current Gruntwork customer (Hint: Even if you're not, you can still join the beta!)
   - How many DevOps engineers does your company have?
   - How many total software engineers does your company have?
   - Why do you want to use Patcher?
   - If Patcher didn't exist, what would you do instead?
   - I realize those are a lot of questions, but that's kind of the point. Let's see if users are in pain enough to get through the questions. Plus, we can make only name and company email required.
- The text we use needs to be more professional. We need to remove phrases like "The Magic of Patcher"

TolaAltalogy:
@josh-padnick Here you can find updated designs of Patcher Landing Page and Patcher Waiting list addressing the feedback above. [Figma](https://www.figma.com/file/RSbGf4YTeeDK2Q331ZFQJM?node-id=1030:75#418348522)

josh-padnick:
Responded in Figma!

TolaAltalogy:
@josh-padnick the Figma is updated. One thing that is missing is the screen from Patcher interface for paragraph "Automatically discover dependencies in your code". I have tagged Jason in the Figma comment and sent him an email about it.

PR #795, opened 2023-06-14
consulting page scroll bug fix

klijakub:
This PR changes: 
- removed the  `consulting.js` file 
- moved consulting scroll function to `main.js` 
- changes one class `box` to `consulting-box` 

@brikis98  It's a weird case because even the build version on the netfliy https://deploy-preview-791--keen-clarke-470db9.netlify.app/ works fine with the scroll
So the problem exists only on the production version. I've moved the function from this separate file to main.js, because there was some error with the import, looks good, but to fully check this we'd have to merge it.

https://github.com/gruntwork-io/gruntwork-io.github.io/assets/4997537/a96bce6b-af29-467f-9bfa-eaf0921e5af1


assets/js/main.js
```diff
@@ -8624,6 +8624,54 @@ $(function () {
   submitButton.on("click", submitForm);
 });
 
+/* Consulting page */
+(function ($) {
```
brikis98, line 8628: How do we ensure that this code _only_ affects the consulting page?

brikis98 (commented):
Thank you for the PR!

Merging the code back into `main.js` seems like a reasonable workaround. I added a few minor questions to the code changes themselves.

brikis98 (approved):
LGTM, thank you!

PR #796, opened 2023-06-20
contact: clarify book a demo page

iangrunt:


pages/contact/_hero.html
```diff
@@ -1,7 +1,8 @@
 <div class="container">
   <div class="row">
     <div class="col-xs-12 text-center">
-      <h1>Book a Gruntwork Demo</h1>
+      <h1 style="font-size: 2.5em; font-weight: 700; color: #ffffff;">Book a Gruntwork Demo</h1>
+      <h2 style="font-size: 1.5em; font-weight: 400; color: #6c757d;">Speak with a real human about your DevOps Foundations!</h2>
```
brikis98, line 5: NIT: on our other pages, we use a plain `<h1>` (with no inline styles) for the heading and `<p class="lead">...</p>` for the sub-heading. Give that a shot!

iangrunt:
@brikis98 This looks much better; I'm not sure what I was doing before when I thought the default styling looked weird. I do like the little break between the form and title though?

brikis98:
> @brikis98 This looks much better; I'm not sure what I was doing before when I thought the default styling looked weird. 

Thanks for fixing!

> I do like the little break between the form and title though?

I find it a bit odd looking, but I guess it's a matter of taste. One thing to consider though: each one-off like this makes it a bit harder to apply styling changes site-wide, which we're likely to do in the very near future. So this might be a place where consistency trumps taste?

iangrunt:
> I find it a bit odd looking, but I guess it's a matter of taste. One thing to consider though: each one-off like this makes it a bit harder to apply styling changes site-wide, which we're likely to do in the very near future. So this might be a place where consistency trumps taste?

@brikis98 agreed, I removed it. I definitely over-simplified after removing some of the fields and content at first but hopefully this is the right balance for now.

brikis98 (approved):
LGTM

PR #811, opened 2023-07-22
Remove deprecated search from IaC Library

josh-padnick:
Now that we have a [Library Reference](https://docs.gruntwork.io/library/reference/), we should remove the current Library search from the marketing site, which is now outdated.

### DIFF
--- a/pages/infrastructure-as-code-library/_library.html
+++ b/pages/infrastructure-as-code-library/_library.html
@@ -3,7 +3,7 @@
   The Infrastructure as Code Library consists of 300+ reusable, battle-tested
   production-grade Terraform modules for AWS.
 </p>
-<ul>
-  <li><a href="https://docs.gruntwork.io/library/overview/">Read the Library Docs</a></li>
-  <li><a href="https://docs.gruntwork.io/library/reference/">Search the Library</a></li>
-</ul>
+<p>
+  <a href="https://docs.gruntwork.io/library/reference/" class="btn btn-primary" role="button" ga-on="click" ga-event-category="CTA Clicks" ga-event-action="Search Library" ga-event-label="User clicked Search Library from marketing website" target="_blank">Search the library</a>
+  <a href="https://docs.gruntwork.io/library/overview/" class="btn" role="button" ga-on="click" ga-event-category="CTA Clicks" ga-event-action="Read Library Docs" ga-event-label="User clicked Read the Library Docs from marketing website" target="_blank">Read the library docs</a>
+</p>
--- a/pages/infrastructure-as-code-library/_library.html
+++ b/pages/infrastructure-as-code-library/_library.html
@@ -1,15 +1,9 @@
 <h2>What's in the Library?</h2>
 <p>
-  The Infrastructure as Code Library consists of 40+ GitHub repos, some open
-  source, some private, each of which contains reusable, battle-tested
-  infrastructure code for AWS, written in Terraform, Go, Bash,
-  and Python. Check out
-  <a
-    href="https://docs.gruntwork.io/library/overview/"
-    target="_blank"
-    >How to use the Gruntwork Infrastructure as Code Library</a
-  >
-  to see how it all works.
+  The Infrastructure as Code Library consists of 300+ reusable, battle-tested
+  production-grade Terraform modules for AWS.
 </p>
-{% include_relative _search.html %} {% include_relative _library-table.html %}
-{% include_relative _no-results.html %} {% include_relative _modals.html %}
+<ul>
+  <li><a href="https://docs.gruntwork.io/library/overview/">Read the Library Docs</a></li>
+  <li><a href="https://docs.gruntwork.io/library/reference/">Search the Library</a></li>
+</ul>


josh-padnick:
@ebeneliason Would you mind reviewing and approving this (unless you have recommended changes)?

josh-padnick:
@brikis98 Thanks for the review!

1. Great point that merely including links really undersells the library.
2. I'm concerned about an iframe because I think the UX is going to be clumsy. Perhaps we should include a screenshot of some modules such as https://github.com/gruntwork-io/terraform-aws-service-catalog/tree/master/modules/networking/vpc and https://docs.gruntwork.io/reference/services/app-orchestration/amazon-eks/? Or perhaps we show a screenshot of https://docs.gruntwork.io/library/reference/?
3. Great point on repo-copier; that's now deprecated and we should remove it and de-commission it.

@klijakub Is this something your team could assist with? Essentially, the ask here is:

1. Include a screenshot of the Library on pages/infrastructure-as-code-library/_library.html, as suggested above.
2. Remove repo-copier links

ebeneliason:
+1 on bolder callouts for docs/library search. Those are selling points! +1 also on avoiding `iframes` here if we can. A screenshot would be okay, but I think it's more compelling to see them in action, so I'd recommend that we put some larger blocks of content there in place of each of those links to really highlight the offering, with an icon or perhaps a screenshot, and big CTA buttons that lead to each on the docs site. It should be totally obvious at a glance that Library == code + docs.

Something along the lines what we have on the [consulting page](https://deploy-preview-811--keen-clarke-470db9.netlify.app/consulting/#our-previous-work) in the "How we can help" and "Our previous work" sections — I'm thinking the general layout of the carousel (without the carousel itself, just two blocks side by side), without the hover content swap (more like the previous work blocks below), and with actual buttons instead of links (more like the hovered carousel blocks). That's a convoluted description, but hopefully it paints the right picture. When I have a moment I can just try mocking something up. Alternatively we could pass this on to Altalogy for a quick turnaround.

josh-padnick:
@ebeneliason You know, I think you should take a stab at an updated Library page and then Altalogy could help us build it out. We're going to need that for the upcoming web revision anyway. :) 

That being said, is this change worth merging as is to at least direct users to the new library reference versus searching on a currently-broken one? We can then separately pursue the improved library change?

If you believe this is worth merging as is, feel free to approve!

brikis98:
> @brikis98 Thanks for the review!
> 
> 1. Great point that merely including links really undersells the library.
> 2. I'm concerned about an iframe because I think the UX is going to be clumsy. 

I agree iframe looks clumsy, but I assumed you were looking for an inexpensive way to make an improvement here, and it seemed like a better option than two easy-to-miss links. 

> Perhaps we should include a screenshot of some modules such as https://github.com/gruntwork-io/terraform-aws-service-catalog/tree/master/modules/networking/vpc and https://docs.gruntwork.io/reference/services/app-orchestration/amazon-eks/? Or perhaps we show a screenshot of https://docs.gruntwork.io/library/reference/?

Screenshot is probably better than two easy-to-miss links as well... But still feels underwhelming. 

> 3. Great point on repo-copier; that's now deprecated and we should remove it and de-commission it.
> 
> @klijakub Is this something your team could assist with? Essentially, the ask here is:
> 
> 1. Include a screenshot of the Library on pages/infrastructure-as-code-library/_library.html, as suggested above.
> 2. Remove repo-copier links

Not repo-copier! But Repo Browser! Those are two separate things. https://gruntwork.io/repos is the Repo Browser and:

1. Inbound links to it should be removed.
2. All its URLs should redirect to the Library Reference in the docs site. We should _not_ merely remove Repo Browser with no replacement, as we DID have inbound links to it in various docs and other content, and we don't want those to 404 now.

josh-padnick:
These are all good points of feedback, but addressing any of them exceed my original appetite for this story, which is at least to fix the status quo, which is actually broken. That is, if you go to https://gruntwork.io/infrastructure-as-code-library/ and search, say, "vpc", nothing shows up.

So perhaps we should merge this as is, and then begin a separate effort to improve this page?

@ebeneliason or @brikis98 If you're on board with that, you can go ahead and merge, approve, and release. We should then start a separate GitHub issue where we can more carefully consider the different options and prioritize the work.

klijakub:
> @klijakub Is this something your team could assist with?

@josh-padnick sorry for being late. Do you still need help with this? If so we can take care of this of course. Please let me know

josh-padnick:
Alright, I'd like to adopt a "bias for action" so I've made a simple change to address Jim's helpful feedback that adds a button for the "search the library" link since I agree it'd be easy to miss. 

At this point, I'm past my appetite for this story and I'd like to get this merged. We can come up with better ways of doing this when we have more time to invest. @ebeneliason Would you mind approving this one so we can merge and release?

@klijakub Thanks for your input; we'll have a bunch of change requests coming to you soon, but no further action is needed on this one.

Thanks all for your input.

image

ebeneliason (approved):
I'll go ahead and approve and merge this per your request. A broken page is a bad look.

With that said, I agree we should prioritize having Altalogy whip up something with a little more presence sooner rather than later. Also, this PR isn't complete given that it doesn't remove all of the defunct JS code and includes, which our best-practices would generally require before merging. A more holistic fix is warranted on both front and back ends. I'll add a tech-debt task in Jira to track the additional back-end cleanup work.

ebeneliason:
Issue filed: https://gruntwork.atlassian.net/browse/CORE-1188

brikis98:
@ebeneliason Should we do a release to roll these changes out?

josh-padnick:
Just did this. No further action required!

ebeneliason:
Thanks for that! I meant to do so yesterday and got sidetracked.

PR #814, opened 2023-08-15
Add a new banner include and show OpenTF blog post banner

ebeneliason:
Here's a preliminary banner include and OpenTF banner design, for feedback.


josh-padnick (approved):
LGTM!

PR #844, opened 2024-03-05
README update for deploys

ZachGoldberg:
Talking to Jim, sounds like the releaser-bot wasn't ever setup for this repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ To deploy the site:
 
 1. Create a PR with your code changes
 1. After the PR has been approved, merge it into `master`
-1. Create a new tag, incrementing the version number using semantic versioning, push the tag up to github
+1. Create a new tag, you can do this manually via git or in the subsequent step on the releases page - be sure to increment the version number using semantic versioning
 1. Go to the [releases page](/releases) and create a draft release with the relevant information (use the "Generate Release Notes" button to make your life easier)
 1. Release it
 1. The CI/CD pipeline will deploy it automatically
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,8 +31,8 @@ To deploy the site:
 
 1. Create a PR with your code changes
 1. After the PR has been approved, merge it into `master`
-1. The releaser bot will create a Draft release
-1. Go to the [releases page](/releases) and update the draft release with the relevant information
+1. Create a new tag, incrementing the version number using semantic versioning, push the tag up to github
+1. Go to the [releases page](/releases) and create a draft release with the relevant information (use the "Generate Release Notes" button to make your life easier)
```
brikis98, line 35: You don't need to create a tag separately. Creating a new release on the releases page will do it automatically.

ZachGoldberg: Curious - when I went to the draft releases page I could only hit the "Generate Release Notes" button if I selected a tag.  I suppose then a tag isn't _strictly_ required, but perhaps is easier than typing up the notes manually?  That sound right?

brikis98: Oh, that might be a GitHub UX issue. In the "Choose a tag" drop down, you can actually type in a new tag, and hit enter, and when you create the release, it'll create the tag too!

ZachGoldberg: Yep, you're right, I had never created a tag before in that way and didn't notice the feature.  Updated the docs to reflect both options.

brikis98 (approved):
LGTM

Resonance1584 (approved):

